PR #42, opened 2014-11-04
implement support for controllers, this helps with #22

wwwdata:
with this pull request, it is possible to addionally register a controller for a resource to do all kinds of stuff, because the controller has access to the request and the struct from the resource


stephenmuss:
Are there any plans to document controllers or to include examples? It would be quite helpful.

PR #65, opened 2015-03-03
latest specs from jsonapi.org

olivoil:
The latest specs for [json api](http://jsonapi.org/format/) state:

* Primary data MUST appear under a top-level key named "data".
* Error objects MUST appear under a top-level key named "errors".
* Linkage MUST be expressed as:
    * type and id members for to-one relationships. type is not required if the value of id is null.
    * type and ids members for homogeneous to-many relationships. type is not required if the value of ids is an empty array ([]).
    * A data member whose value is an array of objects each containing type and id members for heterogeneous to-many relationships.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ slices the values for you.
 Example Request
 GET /people?fields=id,name,age
 
-req["people"] contains values: ["id", "name", "age"]
+req.QueryParams["fields"] contains values: ["id", "name", "age"]
 ```
 
 ### Use Custom Controllers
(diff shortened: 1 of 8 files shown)


wwwdata:
wow, thx for this big PR! Looks really awesome, id's say we can merge this tomorrow if somebody else reviews it too! :+1:

sharpner:
nice ;) awesome. thanks for your commit. 
Just 2 minor things: 
1. Could you remove any uncommented code before we merge this pr?
2. It would be awesome if you could squash and rebase your commits beforehand

wwwdata:
Yes, I agree with @sharpner.

olivoil:
@sharpner @wwwdata done. I removed any commented code and squashed into one commit

PR #68, opened 2015-03-07
Resource links

wwwdata:
This PR introduces the possibility to not specify any related IDs at all and just let a link be auto generated for the client to load related data. See http://jsonapi.org/format/#document-structure-resource-relationships

This introduces the `resource` member in the links object. The resource URL get's automatically generated by Marshalling a struct manually and via the API. For manually marshalling, I introduced 2 more methods to specify a API Prefix. This is now needed because before this feature, the serialisation had nothing to do with URLs. 

So, a new GET Route get's created for referenced structs now like `/v1/posts/1/comments`. If a client requests this URL, the request get's routed to the `FindAll` method of the Resource for `comments` with a Query parameter `PostsID` which in this example has the value 1.

Please let me know what you think about that. 

What I am wondering is whether it is to required to also include the hostname in the `resource` field for the URL. What we now also could easily implement is the `self` field. But I left that out for now, because we do not generate routes for relationships at the moment.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -111,6 +111,8 @@ func (ctx *marshalingContext) marshalStruct(val *reflect.Value, isLinked bool) e
 	idFieldRegex := regexp.MustCompile("^.*ID$")
 
 	valType := val.Type()
+	name := jsonify(pluralize(valType.Name()))
+
 	for i := 0; i < val.NumField(); i++ {
 		tag := valType.Field(i).Tag.Get("json")
 		if tag == "-" {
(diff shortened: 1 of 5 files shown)


marshal.go
```diff
@@ -213,11 +238,18 @@ func (ctx *marshalingContext) marshalStruct(val *reflect.Value, isLinked bool) e
...
 					}
 				} else {
-					linksMap[keyNameWithoutID] = nil
+					linksMap[keyNameWithoutID] = map[string]interface{}{
+						"type":     pluralize(jsonify(structFieldValue.Type().Elem().Name())),
+						"resource": fmt.Sprintf("/%s/%s/%s", name, result["id"], keyNameWithoutID),
+					}
+				}
+				if ctx.prefix != "/" {
+					linksMap[keyNameWithoutID].(map[string]interface{})["resource"] = fmt.Sprintf("%s%s", ctx.prefix, linksMap[keyNameWithoutID].(map[string]interface{})["resource"])
```
sharpner: those casts might fail, shouldn' t we check the "ok" attribute accordingly?

wwwdata: I will change that so that casting is not needed anymore

sharpner: :+1:

Issue #72, opened 2015-03-12
Slash at the end of a URI causes a 307 (Temporary Redirect) response status-code


ben-crowhurst:
curl -XPOST http://localhost:8080/v1/feeds/ -v --data '{ "data":{ "content": "data, here lads", "name": "England weather fault predictions", "description":"corvusoft weather based fault predictor", "supplier": "Corvusoft Solutions"}}'

The additional / at the end of feeds/ causes a HTTP response status of 307 (Temporary Redirect). This has caused a fair bit of confusion over the past hour.   Would recommend 404 (Not Found).

RFC 2616
> 10.3.8 307 Temporary Redirect

> The requested resource resides temporarily under a different URI. Since the redirection MAY be altered
> on occasion, the client SHOULD continue to use the Request-URI for future requests. This response is
> only cacheable if indicated by a Cache-Control or Expires header field.

> The temporary URI SHOULD be given by the Location field in the response. Unless the request method
> was HEAD, the entity of the response SHOULD contain a short hypertext note with a hyperlink to the
> new URI(s) , since many pre-HTTP/1.1 user agents do not understand the 307 status. Therefore, the
> note SHOULD contain the information necessary for a user to repeat the original request on the new
> URI.

> If the 307 status code is received in response to a request other than GET or HEAD, the user agent
> MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might
> change the conditions under which the request was issued.

PR #74, opened 2015-03-12
Introduce method to disable automatic redirects

sharpner:
The default behaviour of api2go will be
307 on routes with trailing / but this feature can be
disabled

This provides the functionality wanted in #72


wwwdata:
:+1:

Issue #71, opened 2015-03-11
Compound Documents show under 'linked' not 'included' as per specification


ben-crowhurst:
Given JSONAPI 1.0 will not be ratified until the end of the month. I'm not sure of the importance of this issue.

http://jsonapi.org/format/#document-structure-compound-documents

wwwdata:
yeah, they changed it from linked to included like 2 days ago

ben-crowhurst:
Ah, sorry to pester.

wwwdata:
No, we needed this ticket anyway! thx :trollface:

PR #92, opened 2015-04-07
Rename fields rc3

### LINKED ISSUE #81: Rename field names
Check all our generated field names and update them for RC3. For example `resource` to `related` etc...

wwwdata:
this resolves #81 and #71 

I looked over jsonapi specs again and did not find any more fields that we have to rename. Apart from renaming we have to continue implementing the other features from the RC3 tickets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,17 +139,33 @@ will yield
       "type": "posts",
       "links": {
         "comments": {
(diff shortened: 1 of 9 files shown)


sharpner:
looks good! But I think the example needs to be updated as well. There is still an "ids" field.

PR #95, opened 2015-04-11
Pagination

### LINKED ISSUE #82: Pagination
Pagination is extremely useful if dealing with a lot of data of the same type. We should support it like described here: http://jsonapi.org/format/#fetching-pagination

Basically use the specified query parameters: 
  * page[number] 
  * page[size]

Or
  * page[offset] 
  * page[limit]

and the generation of pagination links in the `links` object:
  * "first" - the first page of data
  * "last" - the last page of data
  * "prev" - the previous page of data
  * "next" - the next page of data

wwwdata:
This fixes #82


wwwdata:
todo: 
- extend example 
- write Readme

will do that before merging

sharpner:
great :+1:

PR #97, opened 2015-04-14
update readme, example to use PATCH instead of PUT

maxbeatty:
Very small update to the README and example comments to specify using `PATCH` instead of `PUT`. Everything else was great to get up and running :smile_cat:


wwwdata:
thx a lot!

PR #105, opened 2015-04-16
Rename struct fields 

### LINKED ISSUE #91: Replace CamelCase with Dashes
jsonapi now recommends to use dashes for naming of fields

> It is recommended that resource types, attribute names, and association names be "dasherized"; i.e. consist of only lower case alphanumeric characters and dashes (U+002D HYPHEN-MINUS, "-").

> It is also recommended that resource types be pluralized. Dasherized and pluralized resource types can be used as URL segments without translation, as discussed below.

sharpner:
Fixes #91

### DIFF
--- a/jsonapi/helpers.go
+++ b/jsonapi/helpers.go
@@ -1,6 +1,7 @@
 package jsonapi
 
 import (
+	"reflect"
 	"strings"
 	"unicode"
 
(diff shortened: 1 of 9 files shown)


jsonapi/helpers.go
```diff
@@ -79,3 +80,26 @@ func Pluralize(word string) string {
 func Singularize(word string) string {
 	return inflector.Singularize(word)
 }
+
+//GetTagValuesByName returns one api2go setting.
```
wwwdata: please one empty space after slashes

wwwdata:
I would suggest to modify the `jsonapi.Dejsonify` and `jsonapi.Jsonify` methods too, so that they will encode CamelCase to dashes-case, like:

LargeGopher => large-gopher
GopherUUID => gopher-uuid (special case)

but we could also leave that out for now and if a user wants to use dashes, he has to use our tags

wwwdata:
:+1:

sharpner:
I strongly tend to make the magic renaming feature a nice to have for the future, because it is really complex to find a solution that fits every scenario.

wwwdata:
yes, I agree. Let's update the README and then merge it

sharpner:
done

Issue #80, opened 2015-03-17
Routes to manipulate Relationships


wwwdata:
See http://jsonapi.org/format/#crud-updating-to-one-relationships

to one example

```
PATCH /articles/1/links/author
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": { "type": "people", "id": "12" }
}
```

to many example

```
PATCH /articles/1/links/tags
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": [
    { "type": "tags", "id": "2" },
    { "type": "tags", "id": "3" }
  ]
}
```

wwwdata:
3 Methods must be supported now
- PATCH to replace all relationships of a kind
- POST to add more to-many relationships
- DELETE to delete some or all to-many relationships and one to-one

wwwdata:
We need 2 more methods for entity structs to support POST and DELETE, something like
- `AddToManyIDs(name string, IDs []string)`
- `DeleteToManyIDs(name string, IDs []string)`

In addition to that it would also be nice to have explicit set methods for to-one and to-many relationships. So that it is possible to explicitly set something to zero, which means delete a relationship(s) completely.

`SetReferencedIDs(IDs []jsonapi.ReferenceID) ` should be replaced which more concrete methods:
- `SetToOneReferenceID(name string, ID string)`
- `SetToManyReferenceIDs(name string, IDs []string)`

PR #107, opened 2015-04-20
Relationship manipulation

wwwdata:
This implements #80.

The huge brainfuck in this code is again reflections and interfaces. Very often a stupid trick is used to get an addressable value of an `interface{}` by using a slice. Because with a slice it is possible to get a pointer to an `interface{}` type via reflection...

But the feature works

Todos:
- [x] Documentation
- [x] Update example


sharpner:
Awesome!

PR #108, opened 2015-04-20
Method to get instance of julienschmidt httprouter

### LINKED ISSUE #106: Adding custom endpoints/handlers to resources
It's likely that I'm misunderstanding how this package should be used, so it's probably easiest to explain what I'm trying to accomplish since I imagine there is a good reason `API.router` is private.

If I have a userResource that stores a username and passwordHash, like in the crud_example, and I can create new users by sending a POST request to `/v0/users/`, how could I add a custom endpoint with a custom handler, such as `/v0/users/login` and `/v0/users/logout`?

wwwdata:
The idea came up in #106. You can now choose to get a Standard go
Handler or the underlying router instance to further customize api2go
with more routes.

This resolves #106

### DIFF
--- a/api.go
+++ b/api.go
@@ -679,3 +679,8 @@ func handleError(err error, w http.ResponseWriter) {
 func (api *API) Handler() http.Handler {
 	return api.router
 }
+
+// Router can be used instead of Handler() to get the instance of julienschmidt httprouter.
+func (api *API) Router() *httprouter.Router {
+	return api.router
+}


sharpner:
so we want the vendor lock? :D maybe we can open a pull request at julienschmidt so the router exposes an interface?

wwwdata:
You can still use the old method that returns a plain handler. So there is no vendor lock in :P

sharpner:
Well but since we cannot know who might use the router in the future, as soon as we are stable, we're stuck with this method ^^

sharpner:
But I get the point :)

PR #110, opened 2015-04-30
Sql null types

### LINKED ISSUE #109: Marshalling of sql.Null-Types
Thanks to @Ramzeus. api2go currently does not serialize the sql.Null Types correctly. They are marshalled as is.

The problem is that the JSON serialization seems to be "wrong" when using api2go in order to serialize the data to JSON. Here is what it might look like:

```json
{  
   "data":[  
      {  
         "accessControlListId":{  
            "Int64":0,
            "Valid":false
         },
         "description":{  
            "String":"",
            "Valid":true
         },
         "displaySchemeId":{  
            "Int64":0,
            "Valid":false
         },
         "id":"379",
         "inactive":false,
         "name":"DISE22",
         "networkId":1,
         "playlistId":{  
            "Int64":0,
            "Valid":false
         },
         "screenDefinitionId":{  
            "Int64":0,
            "Valid":false
         },
         "settingsGroupId":{  
            "Int64":0,
            "Valid":false
         },
         "type":"players"
      }
   ]
}
```
In this case I think this is what would be preferable:
```json
{  
   "data":[  
      {  
         "accessControlListId":null,
         "description":"",
         "displaySchemeId":null,
         "id":"379",
         "inactive":false,
         "name":"DISE22",
         "networkId":1,
         "playlistId":null,
         "screenDefinitionId":null,
         "settingsGroupId":null,
         "type":"players"
      }
   ]
}
```

wwwdata:
This fixes #109 

However I am not very happy with the current unmarshaling in api2go. For new types it is required to implement some interfaces in order for them to be marshalled/unmarshaled:

- json.Marshaler
- json.Unmarshaler

Marshalling is done right in api2go and the implemented interface will be used. Unmarshaling however cannot currently use implemented Unmarshaler interfaces.

I would suggest that we open another issue for this matter and refactor our unmarshaling.

But SQL-Types work for now. Once we refactored unmarshaling, any custom type that implements the interfaces can be used, that will be very cool :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ go get github.com/univedo/api2go/jsonapi
   - [Unmarshalling with references to other structs](#unmarshalling-with-references-to-other-structs)
 - [Ignoring fields](#ignoring-fields)
 - [Manual marshaling / unmarshaling](#manual-marshaling--unmarshaling)
+- [SQL Null-Types](#sql-null-types)
 - [Building a REST API](#building-a-rest-api)
   - [Query Params](#query-params)
   - [Using Pagination](#using-pagination)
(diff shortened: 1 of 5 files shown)


wwwdata:
I found a very easy solution in my last commit :P 

The interface expects a `[]byte`. The solution was to marshal it to a `[]byte` again and put that into the `UnmarshalJSON` interface method.

wwwdata:
I opened an issue #111 to clean to refactor unmarshaling.

Issue #113, opened 2015-05-03
Unmarshal to single struct


pengux:
Currently jsonapi doesn't seem to allow unmarshalling to a single struct instead of a []UnmarshalIdentifier, but according to the spec:

> The request MUST include a single resource object as primary data

wwwdata:
you are right, this is confusing. We should add the possibility to use a single struct as well.

wwwdata:
we had a slice for it because in a previous specification it was possible to create multiple objects. I will open a PR soon that allows to use a struct directly. I would like to support both because of this idea here #112. From a client perspective (if we will implement that too) it is still required to unmarshal multiple objects.

pengux:
Yes, I agree with supporting both. The standard can be a single object and then we can have a for example UnmarshalManyToJSON() for multiple objects.

Issue #112, opened 2015-05-01
api2go jsonapi client


pmccarren:
It would be awesome if we could use api2go as a golang client, accessing a jsonapi server. Most of the functionality is currently in place, however this seems like it makes sense to do this after the Unmarshal refactorization (#111) project is done.

We would need to add support for Unmarshaling 'Composer Documents': http://jsonapi.org/format/#document-structure-compound-documents

dopin:
+1
I'm very looking forward to this.

sharpner:
good to know! Any special requirements we should keep in mind?

dopin:
Well, I'm a novice at Go, so I don't have any idea but well written documents would help users like me, I guess.
I have JSON API written in Rails and just started to develop a CLI for that.

Oh I think it would be nice that api2go could do like this:

```go
var posts []Post
body, err := ioutil.ReadAll(response.Body) // net/http
api2go.Unmarshal(body, posts)
```

wwwdata:
if you only have simple posts without compound objects, it should already work. Currently only our marshaller supports the generation of a json with compound objects included in it. We have to add this feature to the unmarshaller too.

dopin:
It has `relationships` in it. Is that what you mean by "compound objects"?
I'm trying to unmarshal the data but I get `[]`. :cry: I think I'm doing something wrong but I still don't know the right way.

wwwdata:
as long as you only have relationship IDs, it should work, because that is what works when you use it server-side. Compound objects are objects that you can find under the `included` key in the json. For example if you send a blog post including the comment objects.

dopin:
I'm sorry. I still can't understand.
I wrote code like this. 

```go
type App struct {
	Id			string	`jsonapi:"primary,apps"`
	Name		string	`jsonapi:"attr,name"`
	CreatedAt	time.Time	`jsonapi:"attr,created_at"`
	UpdatedAt	time.Time	`jsonapi:"attr,updated_at"`
	Images		[]*Image	`jsonapi:"relation,images"`
}

type Image struct {
	Id	string				`jsonapi:"-"`
	ImageName	string		`jsonapi:"-"`
	CreatedAt	time.Time	`jsonapi:"-"`
	UpdatedAt	time.Time	`jsonapi:"-"`
	App			App			`jsonapi:"-"`
}

func main() {
    client := &http.Client{}
     req, err := http.NewRequest("GET", "http://localhost:3000/apps", nil)
    resp, err := client.Do(req)
[...]

Issue #111, opened 2015-04-30
Refactor Unmarshaling


wwwdata:
Unmarshaling currently unmarshals into a `map[string]interfaface{}`. 
This is not the most elegant solution, because we have to do some magic about type casting and calling the `json.Unmarshaler` interface. 

We should try to directly scan into the target struct type to avoid all this trouble.

pmccarren:
Do you see support for unmarshmaling 'Compound Documents'? For example, if you are using the jsonapi marshaler from a client's perspective, an api server can return compound documents. (http://jsonapi.org/format/#document-structure-compound-documents)

wwwdata:
Hm. Api2go is currently only for the server. But it's a very good idea I think. So you are having a jsonapi server and want to use a golang program to acces it, right?
It's definitely possible to implement that too. But I would do it after we refactored the current functionality. Maybe we should open a separate ticket for this feature to use api2go as a client.

pmccarren:
Yeah that is correct -- a jsonapi server, and looking to have a golang program access it. Would you like me to open a separate ticket?

wwwdata:
yes. Do it please :) I like the idea

PR #114, opened 2015-05-04
Unmarshal now supports a struct instead of only slice

wwwdata:
Unmarshaling now works with a single struct as well. We used a slice before because in a previous specification it was possible to send multiple objects to the server. 

However, we will clean up the marshalling in ticket #111 

This resolves ticket #113 and also keeps in mind #112. From a clients perspective it would still be required to unmarshal into a slice because a server can send multiple objects inside a json. 

For now both is supported. If there are multiple objects in a json and the target is a struct, the first entry will be unmarshalled into it, and the rest is ignored. After #111 we should check that too and output an error.

### DIFF
--- a/jsonapi/unmarshal.go
+++ b/jsonapi/unmarshal.go
@@ -77,19 +77,37 @@ type EditToManyRelations interface {
 // Unmarshal reads a JSONAPI map to a model struct
 // target must at least implement the `UnmarshalIdentifier` interface.
 func Unmarshal(input map[string]interface{}, target interface{}) error {
+	var (
+		structType reflect.Type
+		sliceVal   reflect.Value
+		isStruct   bool
+	)
+
+	typeError := errors.New("You must pass a pointer to a UnmarshalIdentifier or slice of it to Unmarshal()")
+
 	// Check that target is a *[]Model
 	ptrVal := reflect.ValueOf(target)
 	if ptrVal.Kind() != reflect.Ptr || ptrVal.IsNil() {
-		return errors.New("You must pass a pointer to a []UnmarshalIdentifier to Unmarshal()")
+		return typeError
 	}
-	sliceType := reflect.TypeOf(target).Elem()
-	sliceVal := ptrVal.Elem()
-	if sliceType.Kind() != reflect.Slice {
-		return errors.New("You must pass a pointer to a []UnmarshalIdentifier to Unmarshal()")
+	targetType := reflect.TypeOf(target).Elem()
+
+	if targetType.Kind() != reflect.Slice {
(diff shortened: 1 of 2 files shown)


wwwdata:
Hm, the travis build error seems weird.. The tests work only for tip and 1.4 version. But this makes no sense:

```
3.33s$ ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --trace --race --progress
Failed to run test suite!
	fork/exec /tmp/ginkgo230951662/jsonapi.test: no such file or directory
Ginkgo ran 1 suite in 3.310503106s
Test Suite Failed
```

sharpner:
ginkgo recently updated master, and changed exactly this behaviour. This error message is the bug I got before with symlink folders.

wwwdata:
ok, rebased it on master. ginkgo dropped support for go older than 1.4

trapias:
@wwwdata can I ask a clarification? Does the embedded struct need to satisfy UnmarshalIdentifier too?
I can correctly marshal a struct similar to this (StructA) and its embedded struct (StructB), but I get "could not set field 'StructB' map [...] had wrong type" error when unmarshalling:

	type StructA struct {
		ID    string `json:"-"`
		Field1 string `jsonapi:"name=field-one"`
		FieldB StructB `jsonapi:"name=b-field"`
	}

	func (t StructA) GetID() string {
		return t.ID
	}

	func (t *StructA) SetID(id string) error {
		t.ID = id
		return nil
	}

	type StructB struct {
		A bool `jsonapi:"name=ba-field"`
		B bool `jsonapi:"name=bb-field"`
		C bool `jsonapi:"name=bc-field"`
	}
	
I don't want to set StructB as a relation, shouldn't this be ok to embed my second struct like this?
Thanks,
al.

wwwdata:
hm you want to have an "embedded" object without a relationship. I don't know if this is working with api2go. The jsonapi.org standard discourages you from doing so. Nothing get's embedded, everything has reference IDs. If you want to load a referenced struct within one request, it is in the `included` array.

Marshalling should just work, yes. Unmarshalling will not work because we do a lot of custom type-switching stuff. But maybe if we improve unmarshaling with #111 it will just work too. But apart form that, this is not how you make a clean api with the jsonapi standard.

trapias:
@wwwdata thanks. Since jsonapi tells "Attributes may contain any valid JSON value." (http://jsonapi.org/format/#document-resource-object-attributes) I thought I could do embed such an object without the need of a reference. Will better read jsonapi specs, thanks ;)

sharpner:
you can always ask questions like that on the gitter channel. It's much easier to help you there! ;-) thx for using api2go!

Issue #86, opened 2015-03-23
Add support for extensions


sharpner:
jsonapi.org allows custom extensions. http://jsonapi.org/format/#extending 

We should implement a generic way to register those.

sharpner:
this should be done when #147 gets merged.

PR #119, opened 2015-05-26
Added ContentMarshaler interface

jblazquez:
Hi,

For your consideration here's a proposed new extension point for `api2go` that allows a user of the library to support encodings other than JSON for requests and responses while still being able to take advantage of the autogenerated routes.

The motivation for this change is to allow for higher-performance communication with clients that can speak binary formats that are more optimized than JSON, such as [CBOR](http://tools.ietf.org/html/rfc7049) or [MessagePack](http://msgpack.org/). Coupling these formats with HTTP/2 support can yield significant performance improvements for chatty JSON APIs.

The extension point consists of one new interface and one new `API` constructor:

```go
type ContentMarshaler interface {
  Marshal(i interface{}) ([]byte, error)
  Unmarshal(data []byte, i interface{}) error
}

func NewAPIWithMarshalers(prefix string, baseURL string, marshalers map[string]ContentMarshaler) *API
```

By default the behavior of the library is exactly the same as before: only JSON encoding is supported. However, if the user provides a marshaler for a `Content-Type` other than the standard `application/vnd.api+json` then a client that sends an `Accept` header including that content type will be responded to using that format instead of JSON.

The changes to the library are small because the `jsonapi` subpackage needed no changes at all. It already provided all the necessary support with the general `Marshal` and `MarshalWithURLs` functions. Only the main API implementation had to change a little to pass around some additional context to allow for the right marshaler to be selected.

The existing unit tests provided good test coverage and ensured that no regressions for the default JSON encoding happened. I extended the CRUD example to show how the new extension point could be used.

Thanks.


sharpner:
Hi! Thanks for the pull request. I like the idea of introducing the possibility to exchange ContentMarshalers. 

But one question, since this library is specifically for jsonapi.org would anyone use this with different encodings?

jblazquez:
Yeah, I thought about it. Certainly the intended format of the JSON API spec is JSON, but the spec itself [does have some extensibility hooks](http://jsonapi.org/extensions/#extension-negotiation). One of the paragraphs there reads:

```
For example: a response that includes the header Content-Type: application/vnd.api+json; ext="ext1,ext2"; supported-ext="ext1,ext2,ext3" indicates that the response document is formatted according to the extensions "ext1" and "ext2". Another example: a request that includes the header Content-Type: application/vnd.api+json; ext="ext1,ext2" indicates that the request document is formatted according to the extensions "ext1" and "ext2".
```

One interpretation of that paragraph is that the spec allows formats other than JSON as extensions. For example, it would be conceivable to have a server return responses with `Content-Type: application/vnd.api+json; ext="+cbor"; supported-ext="+cbor,+yaml,+msgpack"` where it's basically saying "this response is structured according to the JSON API spec, and I'm currently using the CBOR format extension, and I support YAML and MessagePack as well".

This might be a stretch and I'd understand if you don't want to introduce this kind of flexibility in the library.

However, the content marshaler concept might still be useful for a user who wants to use JSON but also wants to customize how responses are formatted. For example, they might want to return pretty-printed JSON instead of the default compact JSON, or they might want to use a third-party JSON library for some reason (performance, etc).

api.go
```diff
@@ -798,61 +845,89 @@ func (res *resource) handleDelete(w http.ResponseWriter, r *http.Request, ps htt
...
+}
+
+func selectContentMarshaler(r *http.Request, marshalers map[string]ContentMarshaler) (marshaler ContentMarshaler, contentType string) {
+	if accept, found := r.Header["Accept"]; found {
+		var contentTypes []string
+		for ct := range marshalers {
+			contentTypes = append(contentTypes, ct)
+		}
+
+		contentType = goautoneg.Negotiate(accept[0], contentTypes)
```
sharpner: could we use https://github.com/golang/gddo/blob/master/httputil/negotiate.go for negotiation? goautoeng has no updates since 2012 which is pretty old for a golang library imho

sharpner:
I'm sold. Since this refactoring mostly decouples the implementation it won't hurt. What do you think @wwwdata ?

api.go
jblazquez: Sure, sounds good.

sharpner: awesome!

jblazquez:
OK, pushed a new commit to switch to a more actively-maintained package to do Content-Type negotiation. Also used this opportunity to tweak the wording on the README and the content marshaler example to deemphasize the use case for non-JSON formats and instead emphasize the use case for tweaking the JSON output generated by the library. Let me know if you don't agree with the current wording.

I think the non-JSON format use case is useful but since it depends on JSON API extensions we might want to wait until #86 is solved before we emphasize that too much.

api.go
```diff
@@ -798,61 +845,89 @@ func (res *resource) handleDelete(w http.ResponseWriter, r *http.Request, ps htt
...
+	marshaler, contentType := selectContentMarshaler(r, marshalers)
+	result, err := marshaler.Marshal(resp)
+	if err != nil {
+		return err
+	}
+	writeResult(w, result, status, contentType)
+	return nil
+}
+
+func selectContentMarshaler(r *http.Request, marshalers map[string]ContentMarshaler) (marshaler ContentMarshaler, contentType string) {
```
wwwdata: hm, I think this should get a unit test just to be sure. I really like this PR, very clean and awesome way to add more formats :)

examples/crud_example.go
```diff
@@ -421,8 +421,23 @@ func (c *chocolateResource) Update(obj interface{}, r api2go.Request) error {
...
+}
+
+func (m PrettyJSONContentMarshaler) Unmarshal(data []byte, i interface{}) error {
+	return json.Unmarshal(data, i)
+}
+
 func main() {
-	api := api2go.NewAPIWithBaseURL("v0", "http://localhost:31415")
+	marshalers := map[string]api2go.ContentMarshaler{
+		"application/vnd.api+json": PrettyJSONContentMarshaler{},
```
wwwdata: nice! the example outputs formatted json now.

api.go
jblazquez: You're right, at the very least the `selectContentMarshaler` function could use some tests. Let me get some written.

jblazquez:
@wwwdata @sharpner added a few unit tests to ensure content marshaler selection is sound.

wwwdata:
:+1:

PR #120, opened 2015-05-31
Add support for pointer resources

jblazquez:
This is a proposal to support pointer resources in addition to non-pointer resources. That is, adding support for this:

```go
api.AddResource(&Post{}, &PostsSource{})
```

The use case is being able to easily link resources so that a change to a linked resource is reflected when retrieving the representation of a resource that links to it.

Currently, the following behavior of the pretty simple CRUD example is very surprising (and probably unintended):

```json
# Create a chocolate with the name sweet
$ curl -X POST http://localhost:31415/v0/chocolates -d '{"data" : [{"type" : "chocolates" , "name" : "Ritter Sport", "taste": "Very Good"}]}'
{
    "data": {
        "id": "1",
        "name": "Ritter Sport",
        "taste": "Very Good",
        "type": "chocolates"
    }
}

# Create a user with a sweet
$ curl -X POST http://localhost:31415/v0/users -d '{"data" : [{"type" : "users" , "user-name" : "marvin", "links": {"sweets": {"linkage": [{"type": "chocolates", "id": "1"}]}}}]}'
{
    "data": {
        "id": "1",
        "links": {
            "sweets": {
                "linkage": [
                    {
                        "id": "1",
                        "type": "chocolates"
                    }
                ],
                "related": "http://localhost:31415/v0/users/1/sweets",
                "self": "http://localhost:31415/v0/users/1/links/sweets"
            }
        },
        "type": "users",
        "user-name": "marvin"
    },
    "included": [
        {
            "id": "1",
            "name": "Ritter Sport",
            "taste": "Very Good",
            "type": "chocolates"
        }
    ]
}

# Rename the chocolate
$ curl -X PATCH http://localhost:31415/v0/chocolates/1 -d '{ "data" : {"type" : "chocolates", "name" : "Godiva", "id" : "1"}}'

# List chocolates
$ curl -X GET http://localhost:31415/v0/chocolates/1
{
    "data": {
        "id": "1",
        "name": "Godiva",
        "taste": "Very Good",
        "type": "chocolates"
    }
}

# List users
$ curl -X GET http://localhost:31415/v0/users/1
{
    "data": {
        "id": "1",
        "links": {
            "sweets": {
                "linkage": [
                    {
                        "id": "1",
                        "type": "chocolates"
                    }
                ],
                "related": "http://localhost:31415/v0/users/1/sweets",
                "self": "http://localhost:31415/v0/users/1/links/sweets"
            }
        },
        "type": "users",
        "user-name": "marvin"
    },
    "included": [
        {
            "id": "1",
            "name": "Ritter Sport", <=== OLD NAME
            "taste": "Very Good",
            "type": "chocolates"
        }
    ]
}
```

Notice that after updating the `/chocolates/1` resource the change is not reflected in the `/users/1` response's included resources, even though the user says it's still referencing chocolate ID 1 (there are apparently now two different chocolates with the same ID). This is because in the example the `User` struct stores chocolates by copy (`[]Chocolate`) instead of by reference (`[]*Chocolate`).

Changing the example to have the `User` struct store pointers to chocolates results in (IMO) somewhat unintuitive code because CRUD operations still pass copies of everything around when you want to store pointers:

```go
--- a/examples/crud_example.go
+++ b/examples/crud_example.go
@@ -55,6 +55,6 @@ type User struct {
        //rename the username field to user-name.
-       Username      string      `jsonapi:"name=user-name"`
-       PasswordHash  string      `json:"-"`
-       Chocolates    []Chocolate `json:"-"`
-       ChocolatesIDs []string    `json:"-"`
+       Username      string       `jsonapi:"name=user-name"`
+       PasswordHash  string       `json:"-"`
+       Chocolates    []*Chocolate `json:"-"`
+       ChocolatesIDs []string     `json:"-"`
        exists        bool
@@ -101,3 +101,3 @@ func (u User) GetReferencedStructs() []jsonapi.MarshalIdentifier {
        for key := range u.Chocolates {
-               result = append(result, u.Chocolates[key])
+               result = append(result, *u.Chocolates[key])
        }
@@ -162,3 +162,3 @@ func (c *Chocolate) SetID(id string) error {
 type ChocolateStorage struct {
-       chocolates map[string]Chocolate
+       chocolates map[string]*Chocolate
        idCount    int
@@ -170,3 +170,3 @@ func (s ChocolateStorage) GetAll() []Chocolate {
        for key := range s.chocolates {
-               result = append(result, s.chocolates[key])
+               result = append(result, *s.chocolates[key])
        }
@@ -177,3 +177,3 @@ func (s ChocolateStorage) GetAll() []Chocolate {
 // GetOne tasty chocolate
-func (s ChocolateStorage) GetOne(id string) (Chocolate, error) {
+func (s ChocolateStorage) GetOne(id string) (*Chocolate, error) {
        choc, ok := s.chocolates[id]
@@ -183,3 +183,3 @@ func (s ChocolateStorage) GetOne(id string) (Chocolate, error) {

-       return Chocolate{}, fmt.Errorf("Chocolate for id %s not found", id)
+       return nil, fmt.Errorf("Chocolate for id %s not found", id)
 }
@@ -190,3 +190,3 @@ func (s *ChocolateStorage) Insert(c Chocolate) string {
        c.ID = id
-       s.chocolates[id] = c
+       s.chocolates[id] = &c
        s.idCount++
@@ -212,3 +212,3 @@ func (s *ChocolateStorage) Update(c Chocolate) error {
        }
-       s.chocolates[c.ID] = c
+       *s.chocolates[c.ID] = c

@@ -374,3 +374,3 @@ func (s *userResource) Update(obj interface{}, r api2go.Request) error {
        // check references and get embedded objects, in real world, you would make database queries and check all your references
-       user.Chocolates = []Chocolate{}
+       user.Chocolates = []*Chocolate{}
        for _, chocID := range user.ChocolatesIDs {
@@ -398,3 +398,7 @@ func (c *chocolateResource) FindAll(r api2go.Request) (interface{}, error) {
 func (c *chocolateResource) FindOne(ID string, r api2go.Request) (interface{}, error) {
-       return c.storage.GetOne(ID)
+       choc, err := c.storage.GetOne(ID)
+       if err != nil {
+               return Chocolate{}, err
+       }
+       return *choc, nil
 }
@@ -441,3 +445,3 @@ func main() {
        users := make(map[string]User)
-       chocStorage := ChocolateStorage{chocolates: make(map[string]Chocolate), idCount: 1}
+       chocStorage := ChocolateStorage{chocolates: make(map[string]*Chocolate), idCount: 1}
        api.AddResource(User{}, &userResource{users: users, chocStorage: &chocStorage})
```

Another example of surprising behavior is `FindOne` and `FindAll`, which always return copies of the requested resource. If you make your referring resource (e.g. `User`) retrieve its referenced resources (e.g. `Chocolate`) via those functions what you're storing are pointers to some disconnected copies that `FindOne`/`FindAll` returned, not pointers to the actual Chocolates. You would need to write your own accessors (like `GetOne` in the example) that return pointers to the actual resources you want to link to. My first thought would be to reuse the accessors that I already had.

Yet another reason for having the option to use pointers instead of copies is for performance. If your resources are very large and you have many of them all those copies can start adding up.

So this change makes `api2go` accept resources of type pointer to struct in addition to plain structs. When you register a pointer resource then all the CRUD functions both take and return pointers to your resources. This allows your data sources to store pointers and returns those pointers directly, as well as ensuring that no surprising copies happen anywhere.

The changes boil down to adding a number of conditionals that ensure that pointer resources are treated as pointers all throughout. To verify the new codepaths I essentially made the entire set of API tests run twice, once for struct resources and once for pointer resources.

Let me know what you think.

### DIFF
--- a/examples/crud_example.go
+++ b/examples/crud_example.go
@@ -9,9 +9,9 @@ List users:
 	curl -X GET http://localhost:31415/v0/users
 
 List paginated users:
-	curl -X GET http://localhost:31415/v0/users?page[offset]=0&page[limit]=2
+	curl -X GET 'http://localhost:31415/v0/users?page\[offset\]=0&page\[limit\]=2'
 OR
-	curl -X GET http://localhost:31415/v0/users?page[number]=1&page[size]=2
+	curl -X GET 'http://localhost:31415/v0/users?page\[number\]=1&page\[size\]=2'
 
 Update:
 	curl -vX PATCH http://localhost:31415/v0/users/1 -d '{ "data" : {"type" : "users", "user-name" : "better marvin", "id" : "1"}}'
@@ -53,10 +53,10 @@ import (
 type User struct {
 	ID string
 	//rename the username field to user-name.
-	Username      string      `jsonapi:"name=user-name"`
-	PasswordHash  string      `json:"-"`
-	Chocolates    []Chocolate `json:"-"`
-	ChocolatesIDs []string    `json:"-"`
+	Username      string       `jsonapi:"name=user-name"`
+	PasswordHash  string       `json:"-"`
+	Chocolates    []*Chocolate `json:"-"`
+	ChocolatesIDs []string     `json:"-"`
 	exists        bool
 }
 
@@ -147,7 +147,7 @@ type Chocolate struct {
 }
 
 // GetID to satisfy jsonapi.MarshalIdentifier interface
-func (c Chocolate) GetID() string {
+func (c *Chocolate) GetID() string {
 	return c.ID
 }
 
@@ -160,13 +160,13 @@ func (c *Chocolate) SetID(id string) error {
 // ChocolateStorage stores all of the tasty chocolate, needs to be injected into
 // User and Chocolate Resource. In the real world, you would use a database for that.
 type ChocolateStorage struct {
-	chocolates map[string]Chocolate
+	chocolates map[string]*Chocolate
 	idCount    int
 }
 
 // GetAll of the chocolate
-func (s ChocolateStorage) GetAll() []Chocolate {
-	result := []Chocolate{}
+func (s ChocolateStorage) GetAll() []*Chocolate {
+	result := []*Chocolate{}
 	for key := range s.chocolates {
 		result = append(result, s.chocolates[key])
 	}
@@ -175,17 +175,17 @@ func (s ChocolateStorage) GetAll() []Chocolate {
 }
 
 // GetOne tasty chocolate
-func (s ChocolateStorage) GetOne(id string) (Chocolate, error) {
+func (s ChocolateStorage) GetOne(id string) (*Chocolate, error) {
 	choc, ok := s.chocolates[id]
 	if ok {
 		return choc, nil
 	}
 
-	return Chocolate{}, fmt.Errorf("Chocolate for id %s not found", id)
+	return nil, fmt.Errorf("Chocolate for id %s not found", id)
 }
 
 // Insert a fresh one
-func (s *ChocolateStorage) Insert(c Chocolate) string {
+func (s *ChocolateStorage) Insert(c *Chocolate) string {
 	id := fmt.Sprintf("%d", s.idCount)
 	c.ID = id
 	s.chocolates[id] = c
@@ -205,7 +205,7 @@ func (s *ChocolateStorage) Delete(id string) error {
 }
 
 // Update updates an existing chocolate
-func (s *ChocolateStorage) Update(c Chocolate) error {
+func (s *ChocolateStorage) Update(c *Chocolate) error {
 	_, exists := s.chocolates[c.ID]
 	if !exists {
 		return fmt.Errorf("Chocolate with id %s does not exist", c.ID)
@@ -372,7 +372,7 @@ func (s *userResource) Update(obj interface{}, r api2go.Request) error {
 	}
 
 	// check references and get embedded objects, in real world, you would make database queries and check all your references
-	user.Chocolates = []Chocolate{}
+	user.Chocolates = []*Chocolate{}
 	for _, chocID := range user.ChocolatesIDs {
 		choc, err := s.chocStorage.GetOne(chocID)
 		if err != nil {
@@ -400,7 +400,7 @@ func (c *chocolateResource) FindOne(ID string, r api2go.Request) (interface{}, e
 }
 
 func (c *chocolateResource) Create(obj interface{}, r api2go.Request) (string, error) {
-	choc, ok := obj.(Chocolate)
+	choc, ok := obj.(*Chocolate)
 	if !ok {
 		return "", api2go.NewHTTPError(errors.New("Invalid instance given"), "Invalid instance given", http.StatusBadRequest)
 	}
@@ -413,7 +413,7 @@ func (c *chocolateResource) Delete(id string, r api2go.Request) error {
 }
 
 func (c *chocolateResource) Update(obj interface{}, r api2go.Request) error {
-	choc, ok := obj.(Chocolate)
+	choc, ok := obj.(*Chocolate)
 	if !ok {
 		return api2go.NewHTTPError(errors.New("Invalid instance given"), "Invalid instance given", http.StatusBadRequest)
 	}
@@ -439,9 +439,9 @@ func main() {
 
 	api := api2go.NewAPIWithMarshalers("v0", "http://localhost:31415", marshalers)
 	users := make(map[string]User)
-	chocStorage := ChocolateStorage{chocolates: make(map[string]Chocolate), idCount: 1}
+	chocStorage := ChocolateStorage{chocolates: make(map[string]*Chocolate), idCount: 1}
 	api.AddResource(User{}, &userResource{users: users, chocStorage: &chocStorage})
-	api.AddResource(Chocolate{}, &chocolateResource{storage: &chocStorage})
+	api.AddResource(&Chocolate{}, &chocolateResource{storage: &chocStorage})
 
 	fmt.Println("Listening on :31415")
 	handler := api.Handler().(*httprouter.Router)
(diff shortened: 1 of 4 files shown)


examples/crud_example.go
```diff
@@ -9,9 +9,9 @@ List users:
 	curl -X GET http://localhost:31415/v0/users
 
 List paginated users:
-	curl -X GET http://localhost:31415/v0/users?page[offset]=0&page[limit]=2
+	curl -X GET 'http://localhost:31415/v0/users?page\[offset\]=0&page\[limit\]=2'
```
jblazquez: `bash` and/or `curl` on Linux seem to not like unescaped square brackets in URLs.

wwwdata:
hm, yes the example is wrong and thanks for finding that out. But normally you access a database with the resource structs and they don't store the data directly. If they do it like in our minimal example, the wrong behavior occurs.

So, you should access a referenced data-storage struct from the resource structs and everything is fine and we can also change that in the example, right? 

What other usecases are there where a pointer is needed? Does this not make things more complicated and isn't really that necessary?

jblazquez:
Yes, this adds some complexity to the library implementation (although if you exclude unit tests and example changes it's ~10 targeted changes to `api2go` and only 2 changes to `jsonapi`), but I think it can make life a bit easier for users of the library if they choose to use pointers.

Here's an example of how the current behavior can be a bit surprising for people like me who are new to the library. When you're implementing the `api2go` interfaces you normally have to use non-pointer receivers:

```go
// GetID to satisfy jsonapi.MarshalIdentifier interface
func (u User) GetID() string {
	return u.ID
}

// GetReferencedIDs to satisfy the jsonapi.MarshalLinkedRelations interface
func (u User) GetReferencedIDs() []jsonapi.ReferenceID {
	result := []jsonapi.ReferenceID{}
	for _, chocolate := range u.Chocolates {
		result = append(result, jsonapi.ReferenceID{
			ID:   chocolate.ID,
			Type: "chocolates",
			Name: "sweets",
		})
	}

	return result
}
```

But sometimes you have to use pointer receivers:

```go
// SetID to satisfy jsonapi.UnmarshalIdentifier interface
func (u *User) SetID(id string) error {
	u.ID = id
	return nil
}

// SetToManyReferenceIDs sets the sweets reference IDs and satisfies the jsonapi.UnmarshalToManyRelations interface
func (u *User) SetToManyReferenceIDs(name string, IDs []string) error {
	if name == "sweets" {
		u.ChocolatesIDs = IDs
	}

	return errors.New("There is no to-many relationship with the name " + name)
}
```

Of course when you think about it it's obvious: mutating operations need to use pointer receivers, otherwise modifications would be lost on return from the function. But there's some additional thinking that has to happen here in order to do the right thing.

The problem is worse for optional interfaces. If you tried to use a pointer receiver for `GetID` in order to satisfy `MarshalIdentifer` things would blow up instantly, but if you use a pointer receiver by mistake for `GetReferencedIDs` to try to satisfy `MarshalLinkedRelations` then everything will appear to work except that the library won't pick up your relations because the plain struct type doesn't actually satisfy that interface:

```go
func (u *User) GetReferencedIDs() []jsonapi.ReferenceID {
    // ... this will never get called ...
}
```

On the other hand, if you decide to use pointer resources then everything is consistent, you can just use pointer receivers everywhere (and even if you use non-pointer receivers those types still satisfy the interfaces thanks to Go rules). This does have the benefit of less copying of data, but I think the consistency of receivers is where most of the value for the user of the library lies.

wwwdata:
ok, I see your point. As I understood google Go you should only use pointer receivers when you want to modify the struct with the method your are implementing. Otherwise, it is not necessary to use a pointer receiver. It also makes sense because when you read the function signature without a pointer receiver you instantly know that this function does not modify the struct.

Hm, I read this in the official FAQ https://golang.org/doc/faq#methods_on_values_or_pointers

> By the way, pointer receivers are identical to the situation in Java, although in Java the pointers are hidden under the covers; it's Go's value receivers that are unusual.

> Second is the consideration of efficiency. If the receiver is large, a big struct for instance, it will be much cheaper to use a pointer receiver.

> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. See the section on method sets for details.

> For types such as basic types, slices, and small structs, a value receiver is very cheap so unless the semantics of the method requires a pointer, a value receiver is efficient and clear.

So they say, if we have some methods with a pointer receiver, all of them should use a pointer receiver. This would be the situation with api2gos setter methods. So we should always use a pointer receiver, right?

And it says that if you do not use a pointer receiver, the structs contents will always be copied, which can be a very expensive operation, if the struct is a large struct. If the FAQ says so.

So for the sake of efficiency, it is always advisable to use a pointer receiver, no matter what, as far as I understand now. This has always been confusing to me :stuck_out_tongue:

jblazquez:
Yeah, I think value receivers are not super common in Go, so we probably shouldn't force library users to use them in order for things to work as they expect, but for compatibility reasons it probably makes sense to keep the current behavior of the library available when you call `AddResource` with a plain struct.

If we have support for pointer resources then users can just use pointer receivers everywhere. Or if they want to use value receivers for some operations like `GetID` to emphasize that they don't make changes to the struct then they're free to do so because of [this Go rule](https://golang.org/ref/spec#Method_sets):

> The method set of any other type T consists of all methods declared with receiver type T. The method set of the corresponding pointer type *T is the set of all methods declared with receiver *T or T (that is, it also contains the method set of T)

Basically, if you have a pointer to a struct then both kinds of methods, pointer and non-pointer receivers, can be used to satisfy interfaces. If you have a plain struct then only non-pointer receivers can be used.

Essentially, users have a bit more flexibility if they use pointer resources, and as the FAQ says things are a bit more consistent.

wwwdata:
absolutely, thanks a lot for this PR and the explanation. I will carefully look over the changes tomorrow and merge it then :)

PR #121, opened 2015-06-03
attributes object in data object

### LINKED ISSUE #118: 'reserve attributes in top level of resource object'
I'd like to bring some attention to json-api/json-api#588 and specifically commit json-api/json-api@c1b320f6b0e2d43b61e2079d4666d06f3f63215d. This was recently merged into the spec. I think it warrants a top level struct named something like `Resource`. At this point in time this struct would contain `id`, `type`, `attributes`.

Thoughts?

wwwdata:
This is a necessary change for v 1.0

All struct payload now is in an `attributes` sub-object inside the `data` object of the json.

This resolves #118

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,11 @@ go:
 install: 
   - go get -t -d -v ./...
   - go get github.com/onsi/ginkgo/ginkgo
+  - go get -u github.com/golang/lint/golint
 
 script:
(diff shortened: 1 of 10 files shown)


sharpner:
Could you squash those commits?

wwwdata:
yep, done :)

PR #128, opened 2015-07-01
allow client generated IDs #127

### LINKED ISSUE #127: Client-Generated-Id are allowed by the json-api 1.0, not supported in api2go.
If I use id  (client-generated) for a document during POST request I get a 403 Forbidden error.

As per the spec I interpret that it should be left to the client to generate the id and should be allowed.  Is there a flag which can allow me to set client generated id as it is useful in certain use cases for my server using api2go.

```
Client-Generated IDs

A server MAY accept a client-generated ID along with a request to create a resource. An ID MUST be specified with an id key, the value of which MUST be a universally unique identifier. The client SHOULD use a properly generated and formatted UUID as described in RFC 4122 [RFC4122].

A server MUST return 403 Forbidden in response to an unsupported request to create a resource with a client-generated ID.
```

wwwdata:
This fixes #127 

A user now has to check in a resources `Create` method whether he can accept the ID that a client send or not.

### DIFF
--- a/api.go
+++ b/api.go
@@ -619,19 +619,6 @@ func (res *resource) handleCreate(w http.ResponseWriter, r *http.Request, prefix
 	//TODO create multiple objects not only one.
 	newObj := newObjs.Index(0).Interface()
 
-	checkID, ok := newObj.(jsonapi.MarshalIdentifier)
-	if ok {
-		if checkID.GetID() != "" {
-			err := Error{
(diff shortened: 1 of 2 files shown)


sharpner:
can you remember the reason we disabled this in the first place?

wwwdata:
I don't know anymore maybe we just wanted to disable it first and find a better solution later.

PR #129, opened 2015-07-03
fix a bug with multiple relationships of the same type

wwwdata:
because of a wrong sorting after "type" in marshal.go so me references were
skipped and not correctly marshalled. The unique key that must be used is the
name, not the type because it is possible to have multiple references of the
same type, but with a different name.


wwwdata:
will write a test now too for this behavior, I guess ill do one with only 2 structs and one struct links to the other one with a to-one and 2 to-many relations

sharpner:
awesome!

PR #132, opened 2015-07-05
Improve example

### LINKED ISSUE #131: crud_example.go not unmarshalling request
I have a fresh install of the api2go library, but I am running into issues with the example code provided. Whenever I try to create a resource, the data fields are not showing up in the handlers. As you can see from the return statement below, "user-name" is blank. The same thing happens when using the "Create a chocolate with the name sweet" example curl statement. Placing a print statement for user at the beginning to the Create block returns:
{   [] [] false}

I can't tell if the framework is supposed to unmarshall the request into the User struct or if we have to explicitly use the r api2go.Request parameter. 

```
curl -X POST http://localhost:31415/v0/users -d '{"data" : [{"type" : "users" , "user-name" : "marvin"}]}'
{
    "data": {
        "attributes": {
            "id": "1",
            "user-name": ""
        },
        "id": "1",
        "relationships": {
            "sweets": {
                "data": [],
                "links": {
                    "related": "http://localhost:31415/v0/users/1/sweets",
                    "self": "http://localhost:31415/v0/users/1/relationships/sweets"
                }
            }
        },
        "type": "users"
    }
}
```

wwwdata:
The example is now way better! It's split up into separate files and it has it's own integration test suite.

I also found a bug related to updating relationships which is now fixed and covered by an integration test in the example project.

This resolves #131

### DIFF
--- a/api.go
+++ b/api.go
@@ -705,6 +705,11 @@ func (res *resource) handleUpdate(w http.ResponseWriter, r *http.Request, ps htt
 }
 
 func (res *resource) handleReplaceRelation(w http.ResponseWriter, r *http.Request, ps httprouter.Params, relation jsonapi.Reference) error {
+	var (
+		err     error
+		editObj interface{}
+	)
+
 	oldObj, err := res.source.FindOne(ps.ByName("id"), buildRequest(r))
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


wwwdata:
hm, one test failed one time. I guess this is because a json array has an order, and the order in the test was reversed. I will try to find a workaround for that which ignores the order of the elements in the json array.

sharpner:
awesome! thanks for the work :-) :+1:

wwwdata:
pushed it again with chocolate sorting :) should now always work

PR #133, opened 2015-07-07
bad field error message informs users that field was not expected

wkrause13:


sharpner:
Hey! Thx for the PR. I don't quite get it yet, how does the message improve?
Additionally would yo be so kind to remove the second commit? 
It's simply duplicated :-) thx!

wwwdata:
hm and the message is now wrong. The incoming field was not found in the struct, so the struct was missing it, like the message said or?

sharpner:
@wkrause13 I will close this. We wrote all error messages the way we wrote this one. So either we have to change every message (although they are pretty understandable imho) or none.

PR #135, opened 2015-07-14
new feature to rename a struct

### LINKED ISSUE #134: Hyphens for routes
Currently, the route name for a struct is determined by using `jsonapi.Jsonify` in

https://github.com/manyminds/api2go/blob/219e89b3528e885c1cb63186615ae6681921afed/api.go#L318

For fields of structs we currently enable the re-naming to hyphen style by our custom tag: 

```go
type Comment struct {
  ID   int
  Text string `jsonapi:"name=some-content"`
}
```

I guess we need some mechanism to configure the `jsonapi.Jsonify` and `jsonapi.Dejsonify` methods.

What would be the cleanest way to modify these 2 methods? Is it a good option to define an interface with these 2 methods and inject a custom struct into the package? Something like:

```go
type myCustomGen struct {}

func (m myCustomGen) Jsonify(s string) string {}
func (m myCustomGen) Dejsonify(s string) string {}

jsonapi.SetNameGenerator(&myCustomGen)
```

Or do we have to initialize our api with more configuration parameters, and one of them is this interface with the 2 methods?  

This is important for our release: #130

wwwdata:
the current situation always generated the structs name from it's type. So
a UnicornPost struct would have the generated name `unicornPosts`. But maybe,
a user wants to use a different naming scheme, like `unicorn-posts`. This is
now possible.

Maybe we should change the Jsonify and Dejsonify methods in the future that
they automatically generate *hyphenized* names of everything.

This resolves #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ go get github.com/manyminds/api2go/jsonapi
 ## TOC
 - [Examples](#examples)
 - [Interfaces to implement](#interfaces-to-implement)
+  - [EntityNamer](#entitynamer)
   - [MarshalIdentifier](#marshalidentifier)
   - [UnmarshalIdentifier](#unmarshalidentifier)
   - [Marshalling with References to other structs](#marshalling-with-references-to-other-structs)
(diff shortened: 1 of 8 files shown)


mantenie:
:shipit:

sharpner:
Go naming conventions would suggest that the interface is called EntityNamer instead of EntityName. What do you think?

wwwdata:
alright, i'll change it

Issue #137, opened 2015-07-24
Errors struct missing fields or changed structure


geoffgarside:
I've only recently started looking at JSONAPI, but I think the Errors struct doesn't match whats documented at http://jsonapi.org/format/#errors

I think the `Href` might now be `links: { about: "href" }`, also the `Path` might now be `source: { pointer: "path", parameter: "something" }`.

sharpner:
hey thanks for this ticket. You are right, we need to change this for the 1.0 Release.

geoffgarside:
I've been looking into implementing this and I'm not sure which of these options would be better. Keep `Error` as a (mostly) flat structure adding things like `Links` and `LinksAbout` where client code would mainly set `LinksAbout` and this is wrapped into `Links.About` when marshalling. Alternatively, add just the `Links` and `Source` fields and to fill these in client code would need to set `Links = &api2go.ErrorLinks{About: "url"}`.

The second form doesn't look as nice, but as the fields have to be optional theres not really a nice way to do this. The json encoder doesn't support calling a function on a field to ask if its empty or not.

The first option, with the mostly flat structure, would look like this.
```go
type Error struct {
	ID              string       `json:"id,omitempty"`
	Links           *ErrorLinks  `json:"links,omitempty"`
	LinksAbout      string       `json:"-"`
	Status          string       `json:"status,omitempty"`
	Code            string       `json:"code,omitempty"`
	Title           string       `json:"title,omitempty"`
	Detail          string       `json:"detail,omitempty"`
	Source          *ErrorSource `json:"source,omitempty"`
	SourcePointer   string       `json:"-"`
	SourceParameter string       `json:"-"`

	// Deprecated Fields
	Href string `json:"-"` // Please use LinksAbout
	Path string `json:"-"` // Please use Source/Pointer
}
```

The second option looks like this
```go
type Error struct {
	ID              string       `json:"id,omitempty"`
	Links           *ErrorLinks  `json:"links,omitempty"`
	Status          string       `json:"status,omitempty"`
	Code            string       `json:"code,omitempty"`
	Title           string       `json:"title,omitempty"`
	Detail          string       `json:"detail,omitempty"`
	Source          *ErrorSource `json:"source,omitempty"`

	// Deprecated Fields
	Href string `json:"-"` // Please use Links/About
	Path string `json:"-"` // Please use Source/Pointer
}
```

The `ErrorLinks` and `ErrorSource` being the following

```go
type ErrorLinks struct {
[...]

PR #139, opened 2015-07-28
Update api2go.Error structure to json api 1.0 spec

geoffgarside:
The Href and Path fields have been removed. The Links, Source and Meta fields have been added.

To update existing code the Href should be changed to

    Links: api2go.ErrorLinks{About: "Href Value"},

the Path should be changed to

    Source: api2go.ErrorSource{Pointer: "Path Value"},

Refs #137 

`go test` is currently passing.

### DIFF
--- a/error.go
+++ b/error.go
@@ -19,18 +19,39 @@ type HTTPError struct {
 //other semantical application problems
 //for more information see http://jsonapi.org/format/#errors
 type Error struct {
-	ID     string `json:"id,omitempty"`
-	Href   string `json:"href,omitempty"`
-	Status string `json:"status,omitempty"`
-	Code   string `json:"code,omitempty"`
-	Title  string `json:"title,omitempty"`
-	Detail string `json:"detail,omitempty"`
-	Path   string `json:"path,omitempty"`
+	ID     string       `json:"id,omitempty"`
+	Links  *ErrorLinks  `json:"links,omitempty"`
+	Status string       `json:"status,omitempty"`
+	Code   string       `json:"code,omitempty"`
+	Title  string       `json:"title,omitempty"`
+	Detail string       `json:"detail,omitempty"`
+	Source *ErrorSource `json:"source,omitempty"`
+	Meta   interface{}  `json:"meta,omitempty"`
 }
 
 // GetID returns the ID
 func (e Error) GetID() string {
 	return e.ID
+}
+
+//ErrorLinks is used to provide an About URL that leads to
+//further details about the particular occurrence of the problem.
+//
+//for more information see http://jsonapi.org/format/#error-objects
+type ErrorLinks struct {
+	About string `json:"about,omitempty"`
+}
+
+//ErrorSource is used to provide references to the source of an error.
+//
+//The Pointer is a JSON Pointer to the associated entity in the request
+//document.
+//The Paramter is a string indicating which query parameter caused the error.
+//
+//for more information see http://jsonapi.org/format/#error-objects
+type ErrorSource struct {
+	Pointer   string `json:"pointer,omitempty"`
+	Parameter string `json:"parameter,omitempty"`
 }
 
 //marshalError marshals all error types
--- a/api_test.go
+++ b/api_test.go
@@ -281,7 +281,7 @@ func (s *fixtureSource) Create(obj interface{}, req Request) (string, error) {
 
 	if p.Title == "" {
 		err := NewHTTPError(errors.New("Bad request."), "Bad Request", http.StatusBadRequest)
-		err.Errors = append(err.Errors, Error{ID: "SomeErrorID", Path: "Title"})
+		err.Errors = append(err.Errors, Error{ID: "SomeErrorID", Source: &ErrorSource{Pointer: "Title"}})
 		return "", err
 	}
 
@@ -620,7 +620,7 @@ var _ = Describe("RestHandler", func() {
 			Expect(rec.Code).To(Equal(http.StatusOK))
 			Expect(rec.Body.Bytes()).To(MatchJSON(`
 				{"data": {
-					"id": "1", 
+					"id": "1",
 					"type": "users",
 					"attributes": {
 						"name": "Dieter"
@@ -635,7 +635,7 @@ var _ = Describe("RestHandler", func() {
 			Expect(rec.Code).To(Equal(http.StatusOK))
 			Expect(rec.Body.Bytes()).To(MatchJSON(`
 				{"data": [{
-					"id": "1", 
+					"id": "1",
 					"type": "comments",
 					"attributes": {
 						"value": "This is a stupid post!"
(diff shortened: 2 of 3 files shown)


error.go
```diff
@@ -33,6 +36,98 @@ func (e Error) GetID() string {
...
+
+	// Generate our JSON
+	var buf bytes.Buffer
+	if _, err := buf.WriteString("{"); err != nil {
+		return nil, err
+	}
+
+	first := true
+	for _, field := range fields {
+		if first {
```
wwwdata: first is not really needed in the loop, right? It is just set to false.

wwwdata:
looks good :+1:

error.go
```diff
@@ -33,6 +36,98 @@ func (e Error) GetID() string {
...
+	}
+
+	// Generate our JSON
+	var buf bytes.Buffer
+	if _, err := buf.WriteString("{"); err != nil {
+		return nil, err
+	}
+
+	first := true
+	for _, field := range fields {
```
wwwdata: I guess you could also use https://godoc.org/strings#Join here to avoid the manual loop and first check for the separator.

sharpner: I would recommend using the normal json.Marshal method instead of manually building a string:

```
package main

import (
  "encoding/json"
  "fmt"
  "log"
)

//Error struct tbd
type Error struct {
  ID     string      `json:"id,omitempty"`
  Links  ErrorLinks  `json:"links,omitempty"`
  Status string      `json:"status,omitempty"`
  Code   string      `json:"code,omitempty"`
  Title  string      `json:"title,omitempty"`
  Detail string      `json:"detail,omitempty"`
  Source ErrorSource `json:"source,omitempty"`
  Meta   interface{} `json:"meta,omitempty"`
}

//ErrorLinks struct tbd
type ErrorLinks struct {
  About string `json:"about,omitempty"`
}

//MarshalJSON overwrites the default bla
func (e ErrorLinks) MarshalJSON() ([]byte, error) {
  if e.About == "" {
    return []byte(`""`), nil
  }

  return json.Marshal(e.About)
}

// ErrorSource is used bla
type ErrorSource struct {
  Pointer   string `json:"pointer,omitempty"`
  Parameter string `json:"parameter,omitempty"`
}

func main() {
  links := ErrorLinks{}
  source := ErrorSource{}
  e := Error{ID: "id", Links: links, Status: "fisch", Source: source}
  result, err := json.Marshal(e)
  if err != nil {
    log.Fatal(err)
  }

  fmt.Printf("%#v", string(result))

}

```

Additionally please describe public structs to the linting "standard"  e.g.:

//Error struct can be used to......

wwwdata: I checked out your example @sharpner it seems that the normal json marshaller skips nil references completely.. so yes, we do not need to implement the marshaler, right?

Example code

```go
package main

import (
  "encoding/json"
  "fmt"
  "log"
)

//Error struct tbd
type Error struct {
  ID     string      `json:"id,omitempty"`
  Links  *ErrorLinks  `json:"links,omitempty"`
  Status string      `json:"status,omitempty"`
  Code   string      `json:"code,omitempty"`
  Title  string      `json:"title,omitempty"`
  Detail string      `json:"detail,omitempty"`
  Source *ErrorSource `json:"source,omitempty"`
  Meta   interface{} `json:"meta,omitempty"`
}

//ErrorLinks struct tbd
type ErrorLinks struct {
  About string `json:"about,omitempty"`
}

//MarshalJSON overwrites the default bla
func (e ErrorLinks) MarshalJSON() ([]byte, error) {
  if e.About == "" {
    return []byte(`""`), nil
  }

  return json.Marshal(e.About)
}

// ErrorSource is used bla
type ErrorSource struct {
  Pointer   string `json:"pointer,omitempty"`
  Parameter string `json:"parameter,omitempty"`
}

func main() {
  e := Error{ID: "id", Status: "fisch"}
  result, err := json.Marshal(e)
  if err != nil {
    log.Fatal(err)
  }

  fmt.Printf("%#v", string(result))

}
```

sharpner: you were a little bit faster ;) I misunderstood the requirement that omitempty "didn't work". 

But at least that way there is no "manual" encoding. That is always the wrong thing to do :D

geoffgarside:
It depends how you want the structure to look. If you use pointers for the Links and Source fields then you don't need to implement MarshalJSON as the fields can just be set to `nil` and left empty. If you don't use pointers then you need to explicitly implement MarshalJSON in order to handle possible emptiness of Links and Source.

If you don't use pointers then you can just create an `Error` instance and then set `error.Links.About` if you wanted to. Otherwise you need to set `error.Links = &ErrorLinks{About:"value"}`, it depends on how you want the API to go really.

If you implement `MarshalJSON` as you've shown for `ErrorLinks` you'll end up with `"links":"value"` rather than `"links":{"about":"value"}`

I'll have a go at just implementing the MarshalJSON for the ErrorLinks and ErrorSource

sharpner:
its ok for me to do 
```go
error.Links = &ErrorLinks{About:"value"}
```
we could also provide an utlity function like
```go
NewError() *Error
```
which initialies sub pointers if you like. 

I'm just happy if there is no manual json string concat ;)

geoffgarside:
Ok, I can change it to pointers and add a creator method. I've checked further and doing `MarshalJSON` on the ErrorLinks and ErrorSource structs won't let you make them optional as the field name has already been written by that point so even if you return an empty byte array or an empty string the field will still be written into the JSON and then you'll get an error about premature end of JSON :(

geoffgarside:
Yeah I prefer not having to manually write JSON, I was worried about escaping the string values, its possible but the `json.HTMLEncode` function expects a bytes.Buffer to work with and the code looked horrible. The other idea I'd had was an unexported type just for JSON marshalling with pointers for Links and Source fields.

sharpner:
Yeah I tried that as well. It kinda feels easy while reading the docs but you are correct. MarshalJSON for one field does not let you "skip" it. It kinda sucks that omitempty allows no implementation of a custom interface ,)

geoffgarside:
Yes, a nice JSONEmpty() function would be very handy.

geoffgarside:
I think its going to best to leave out the `NewError() *Error` function. If we initialised it with Links and Source fields then client code wouldn't necessarily know to nil them if it wasn't going to use them. I'm presuming that if an "about" value isn't present in the jsonapi errors object then the "links" field shouldn't be there either, i.e. we shouldn't be marshalling a result with `"links":{}` in it.

geoffgarside:
I've pushed updates converting to pointers and adding some documentation for ErrorLinks and ErrorSource. If you're happy with that let me know and I can squash the commits so they'll go in as one commit if you merge the PR.

sharpner:
nice looks much cleaner now! 

I think you can squash them and we are good to go. 

A test with empty links and/or empty source would be bonus :P

geoffgarside:
Squashed and bonused :)

sharpner:
Awesome, I'm more than happy ;) thanks for the nice work!

wwwdata:
:+1: thanks

PR #140, opened 2015-07-28
Replace slack notifications with gitter.

sharpner:
Since we do not use slack anymore we could change notifications to our
public gitter chat.


wwwdata:
Yeah.

PR #144, opened 2015-07-31
add return object to method not allowed error.

### LINKED ISSUE #143: Non-json error responses for invalid path
if I make a request for eg. 
curl -X PATCH http://lcoalhost:3000/sweets 

I get a 405 Method not Allowed which is correct but error object is "Content-Type: text/plain; charset=utf-8" body "Method Not Allowed"

I think any error or responses by the server implmented with api2go should return a jsonapi error object.

sharpner:
Fixes #143 

This commit introduces our custom method not allowed handler
currently its private, since there is no need to replace it from
the outside.

TODO: 
- [x] ContentMarshaler must not be hardcoded.
- [x] Return correct json content type.
- [x] Assure that all errors will always return an error object.

### DIFF
--- a/api.go
+++ b/api.go
@@ -17,6 +17,8 @@ import (
 	"github.com/manyminds/api2go/jsonapi"
 )
 
+const defaultContentTypHeader = "application/vnd.api+json"
+
 // The CRUD interface MUST be implemented in order to use the api2go api.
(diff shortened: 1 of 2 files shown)


sharpner:
@wwwdata I think I'm done, you can review it now ;)

PR #146, opened 2015-08-01
Introduce public MarshalError for content type.

### LINKED ISSUE #117: Make marshalError() and marshalHTTPError() public
When creating custom endpoint handlers, it is necessary to produce custom errors too. So it would be nice if marshalError and marshalHTTPError are public so it can be used by the consuming package. Maybe those methods should be moved to the jsonapi package too?

sharpner:
fixes #117.

The MarshalError method is now public api of the ContentMarshaler interface. It only makes sense here since its necessary to implement different behaviour when using another encoding e.g. cbor.

This PR is a breaking change if someone has already implemented custom content marshalers.

### DIFF
--- a/api_interfaces.go
+++ b/api_interfaces.go
@@ -23,6 +23,7 @@ type CRUD interface {
 type ContentMarshaler interface {
 	Marshal(i interface{}) ([]byte, error)
 	Unmarshal(data []byte, i interface{}) error
+	MarshalError(error) string
 }
 
 // The PaginatedFindAll interface can be optionally implemented to fetch a subset of all records.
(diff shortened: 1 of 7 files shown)


api_interfaces.go
```diff
@@ -23,6 +23,7 @@ type CRUD interface {
 type ContentMarshaler interface {
 	Marshal(i interface{}) ([]byte, error)
 	Unmarshal(data []byte, i interface{}) error
+	MarshalError(error) string
```
sharpner: should this interface enforce the HTTPError type?

wwwdata:
cool! I think you accidentaly added a binary file to your commit

sharpner:
lol! removed it ;) nice catch.

PR #147, opened 2015-08-02
Add return code to Resource interface

### LINKED ISSUE #86: Add support for extensions

### LINKED ISSUE #138: Handle update with payload changes
The specification says that if you're changing any of the properties during the update process, such as computing a sha or setting an `updated-at` timestamp that you need to respond with a 200 OK and return the resource information to the client. At the moment the `handleUpdate` function is only able to respond with a `204 No Content` response.

Is it possible to match this requirement with the current API?

wwwdata:
Create, Update and Delete methods now support an additional return code. In
case of an error, the error return type with a 400 or 500 return code must be
used. In case of success, the int return code can be used.

Whenever an updated response must be sent, the FindOne method is automatically called by the API. 

There is one thing where I do not know how to solve it best, it's the 200 OK Status for Delete. It should be possible to only return some meta data here. Any suggestions?

Resolves #138 
Resolves #86 

Todo:
- [x] Always include meta struct
- [x] Update Readme
- [x] Refactor switch with status codes

### DIFF
--- a/api.go
+++ b/api.go
@@ -831,8 +831,11 @@ func (res *resource) handleDelete(w http.ResponseWriter, r *http.Request, ps htt
 
 	switch response.StatusCode() {
 	case http.StatusOK:
-		// TODO: implement this.. we need meta data as return value
-		return fmt.Errorf("status 200 OK is currently not implemented for Delete methods")
+		data := map[string]interface{}{
+			"meta": response.Metadata(),
+		}
+
+		return marshalResponse(data, w, http.StatusOK, r, res.marshalers)
 	case http.StatusAccepted:
 		w.WriteHeader(http.StatusAccepted)
 		return nil
--- a/api.go
+++ b/api.go
@@ -18,6 +18,24 @@ import (
 
 const defaultContentTypHeader = "application/vnd.api+json"
 
+type response struct {
+	Meta   map[string]interface{}
+	Data   interface{}
+	Status int
+}
+
+func (r response) Metadata() map[string]interface{} {
+	return r.Meta
+}
+
+func (r response) Result() interface{} {
+	return r.Data
+}
+
+func (r response) StatusCode() int {
+	return r.Status
+}
+
 type information struct {
 	prefix  string
 	baseURL string
@@ -339,7 +357,7 @@ func (res *resource) handleIndex(w http.ResponseWriter, r *http.Request, info in
 			return err
 		}
 
-		return respondWithPagination(response.Result(), info, http.StatusOK, paginationLinks, w, r, res.marshalers)
+		return respondWithPagination(response, info, http.StatusOK, paginationLinks, w, r, res.marshalers)
 	}
 	source, ok := res.source.(FindAll)
 	if !ok {
@@ -351,7 +369,7 @@ func (res *resource) handleIndex(w http.ResponseWriter, r *http.Request, info in
 		return err
 	}
 
-	return respondWith(response.Result(), info, http.StatusOK, w, r, res.marshalers)
+	return respondWith(response, info, http.StatusOK, w, r, res.marshalers)
 }
 
 func (res *resource) handleRead(w http.ResponseWriter, r *http.Request, ps httprouter.Params, info information) error {
@@ -363,7 +381,7 @@ func (res *resource) handleRead(w http.ResponseWriter, r *http.Request, ps httpr
 		return err
 	}
 
-	return respondWith(response.Result(), info, http.StatusOK, w, r, res.marshalers)
+	return respondWith(response, info, http.StatusOK, w, r, res.marshalers)
 }
 
 func (res *resource) handleReadRelation(w http.ResponseWriter, r *http.Request, ps httprouter.Params, info information, relation jsonapi.Reference) error {
@@ -445,7 +463,7 @@ func (res *resource) handleLinked(api *API, w http.ResponseWriter, r *http.Reque
 					return err
 				}
 
-				return respondWithPagination(response.Result(), info, http.StatusOK, paginationLinks, w, r, res.marshalers)
+				return respondWithPagination(response, info, http.StatusOK, paginationLinks, w, r, res.marshalers)
 			}
 
 			source, ok := resource.source.(FindAll)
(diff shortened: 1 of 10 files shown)


api.go
```diff
@@ -758,12 +788,24 @@ func getPointerToStruct(oldObj interface{}) interface{} {
...
+	status, err := res.source.Delete(ps.ByName("id"), buildRequest(r))
 	if err != nil {
 		return err
 	}
-	w.WriteHeader(http.StatusNoContent)
-	return nil
+
+	switch status {
+	case http.StatusOK:
+		// TODO: implement this.. we need meta data as return value
```
sharpner: As discussed, we introduce an api2go.Response object =)

wwwdata:
Hm, so this response object that we use. What it would need is
- Response Body, for example for FindOne FindAll Methods
- Status Code
- Potential Error
- And we sometimes only return an ID, so an ID field?

sharpner:
I would keep error separated. Don't you think?

sharpner:
Alternatively how about 
```go
type Responder interface {
  func Metadata() map[string]interface{}
  func StatusCode() int
}
```

wwwdata:
hmm, and what about actual data for FindOne and FindAll methods? We should put that in too, right?

Since it has no real functionality an interface is a bid overkill I think. A simple struct with fields is enough.

sharpner:
since we make it public an interface is always the better solution imho.

wwwdata:
ok, I'll do it as an interface then. Regarding the methods in this interface. Do we just want MetaData and StatusCode or do we want to put more into it? I agree with you that the error should be kept as a separate return value. 

For the Find and FindAll methods, this means we have 3 return values: `interface{}, Responder, error`
All update methods would only have `Responder, error`

The Create method would have these 3 return values: `string, Responder, error`

The Meta object is no big deal, it's allowed to use everything, see: http://jsonapi.org/format/#document-meta

> Where specified, a meta member can be used to include non-standard meta-information. The value of each meta member MUST be an object (a "meta object").

> Any members MAY be specified within meta objects.

> For example:
> ```
{
  "meta": {
    "copyright": "Copyright 2015 Example Corp.",
    "authors": [
      "Yehuda Katz",
      "Steve Klabnik",
      "Dan Gebhardt",
      "Tyler Kellen"
    ]
  },
  "data": {
    // ...
  }
}
```

I don't know if it would be a good thing to also have the payload and generated ID in responder. The downside of that would be that it's not needed for all methods, therefore it is confusing, and better to use more return parameters for these resource methods, right?

**Comment by wwwdata** [comment]
The worst method would be this:

```go
type PaginatedFindAll interface {
	PaginatedFindAll(req Request) (obj interface{}, totalCount uint, Responder, err error)
}
```

sharpner:
Ok I Agree, that's kinda ugly. 

How about this, for PaginatedFindAll the return values are (prototyped :P):
```go
type ResponderList interface {
  Responder
  Results() interface{}
  ResultCount() uint
}
```

```go
type PaginatedFindAll interface {
 PaginatedFindAll(req Request) (ResponderList, error)
}
```

sharpner:
But I'm not really happy with that solution either, Metadata is optional -.- Which leads to Responder only containing StatusCode and Metadata is another optional interface... any ideas?

wwwdata:
another alternative would be to put it all into the same `Responder`. The api2go api for example could benefit from that for the `Update` method. If `Results()` returns something, there is no more need to call `FindOne()` after Updating something, because the `Update()` method directly returns the updated struct. 

This is also useful for the `Create()` method which has to return the struct too if it changed something, like adding a timestamp.

The only thing that would only be needed in one Resource is the `ResultCount()`. So I would leave that out it let it be a separate return value. 

So this would be the new Responder

```go
type Responder interface {
  Metadata() map[string]interface{}
  Result() interface{}
  StatusCode() int
}
```

sharpner:
How would you handle the interface{} of the result? There must be explicit typecast to know if its either one result, or many?

For single results its no problem, 
api2go.MarshalIdentifier is enough, but for multiple results?

wwwdata:
You are right, that's why we always return an []MarshalIdentifier :smile:

sharpner:
seems reasonable :D

wwwdata:
todo:

- return meta informations, refactor respondWith for that case is best I guess
- return codes for relationship updates

api.go
```diff
@@ -493,18 +492,31 @@ func (res *resource) handleCreate(w http.ResponseWriter, r *http.Request, prefix
...
-	if err != nil {
-		return err
-	}
+	result, ok := response.Result().(jsonapi.MarshalIdentifier)
 
-	return respondWith(obj, info, http.StatusCreated, w, r, res.marshalers)
+	if !ok {
+		return fmt.Errorf("Expected one newly created object by resource %s", res.name)
+	}
+	w.Header().Set("Location", prefix+res.name+"/"+result.GetID())
```
sharpner: y set a location header here?

wwwdata: because it's in the spec. The diff is wrong it was already there before

api.go
```diff
@@ -571,11 +583,36 @@ func (res *resource) handleUpdate(w http.ResponseWriter, r *http.Request, ps htt
...
-	if err := res.source.Update(updatingObj, buildRequest(r)); err != nil {
+	response, err := res.source.Update(updatingObj, buildRequest(r))
+
+	if err != nil {
 		return err
 	}
-	w.WriteHeader(http.StatusNoContent)
-	return nil
+
+	switch response.StatusCode() {
```
sharpner: can we move those three switch response.StatusCode() statements into one function?

api.go
sharpner: ah ok :D looked strange, I only know them on redirects. thx

wwwdata:
im back now from vacation and will finish this ok @sharpner?

sharpner:
awesome :D

theduke:
Perfect, than you for the wor!

theduke:
Actually, I have a suggestion:

supply a default Result type that implementes Responder.

wwwdata:
ok, that can be useful. Do you want to implement one and open a PR?

theduke:
Yeah.
I also did quite a bit of refactoring to make the code cleaner.
I'll open a PR in a few days.

PR #151, opened 2015-08-06
Add a QueryParam with the Name of the Reference

dfeyer:
When the API call the findAll method it can be useful to know the Reference Name. If you have a single model with multiple property linked to the same reference type (per ex. a model People, can be linked as Director and Producer in a Film).


wwwdata:
yes that makes sense. In that case, the People FindAll Method would become the `name` for a Movie for both cases and does not know which relationship it has to handle. Know it is clear and your new query parameter would be `moviesName: producer` or `moviesName: director`

thanks

PR #164, opened 2015-08-29
Refactoring and new features.

theduke:
This commit contains quite a few changes:

*) Refactored request handling to be more centralized and clean

With the "getMiddleware()" method I centralized the request building and json marshaling.
This cleans up the code a lot in my opinion.
It was also neccesary for the next change.

Also, the request now contains a Data and a RawData field so both the handle functions and 
the resource handlers can have access to the raw parsed request data.
I also needed this in some cases.

*) added BeforeHandle and AfterHandle hooks

The struct given to RegisterResource() can now contain BeforeHandle() and AfterHandle() methods which will be called before and after the request. I added the neccessary interface.

I also needed this for one of my applications.

*) now including error code in response

My client apps often have to handle specific errors that occur automatically.
That's why I really need the "code" attribute of errors to get to the client.
I think one should just manually log the error that should not go to the client and 
set the code to something the client can use. 

*) added Meta metadata to request object.

Now parsing metadata from the request and made it available in the request struct.

*) changed QueryParams in request to be of type Url.Values (supports .Get())

The query params are of type Url.Values anyway in the original request, and I think it's quite a bit easier to work with the .Get() method.

*) Updated tests

I updated all tests. They are all passing fine for me.

### DIFF
--- a/api.go
+++ b/api.go
@@ -9,7 +9,6 @@ import (
 	"net/url"
 	"reflect"
 	"strconv"
-	"strings"
 
 	"github.com/golang/gddo/httputil"
 	"github.com/julienschmidt/httprouter"
@@ -21,7 +20,10 @@ const defaultContentTypHeader = "application/vnd.api+json"
 type response struct {
 	Meta   map[string]interface{}
 	Data   interface{}
+	RawData []byte
+	Error error
 	Status int
+	Header http.Header
 }
 
 func (r response) Metadata() map[string]interface{} {
@@ -53,14 +55,15 @@ type paginationQueryParams struct {
 	number, size, offset, limit string
 }
 
-func newPaginationQueryParams(r *http.Request) paginationQueryParams {
+func newPaginationQueryParams(r Request) paginationQueryParams {
 	var result paginationQueryParams
 
-	queryParams := r.URL.Query()
-	result.number = queryParams.Get("page[number]")
-	result.size = queryParams.Get("page[size]")
-	result.offset = queryParams.Get("page[offset]")
-	result.limit = queryParams.Get("page[limit]")
+	params := r.QueryParams
+
+	result.number = params.Get("page[number]")
+	result.size = params.Get("page[size]")
+	result.offset = params.Get("page[offset]")
+	result.limit = params.Get("page[limit]")
 
 	return result
 }
@@ -81,16 +84,17 @@ func (p paginationQueryParams) isValid() bool {
 	return false
 }
 
-func (p paginationQueryParams) getLinks(r *http.Request, count uint, info information) (result map[string]string, err error) {
+func (p paginationQueryParams) getLinks(r Request, count uint, info information) (result map[string]string, err error) {
 	result = make(map[string]string)
 
-	params := r.URL.Query()
+	params := r.QueryParams
 	prefix := ""
+
 	baseURL := info.GetBaseURL()
 	if baseURL != "" {
 		prefix = baseURL
 	}
-	requestURL := fmt.Sprintf("%s%s", prefix, r.URL.Path)
+	requestURL := fmt.Sprintf("%s%s", prefix, r.PlainRequest.URL.Path)
 
 	if p.number != "" {
 		// we have number & size params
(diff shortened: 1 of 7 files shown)


wwwdata:
hi, thanks a lot so far, I will now look over the changes. It seems that you have to use gofmt tool for your sources and that you forgot to update the examples in order for all the travis tests to pass.

api.go
```diff
@@ -185,12 +189,108 @@ func (n notAllowedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	resourceType reflect.Type
-	source       CRUD
-	name         string
-	marshalers   map[string]ContentMarshaler
+	resourceType   reflect.Type
+	source         CRUD
+	hooks          CRUDHooks
+	name           string
+	marshalers     map[string]ContentMarshaler
+
```
wwwdata: please remove this blank line

api.go
```diff
@@ -185,12 +189,108 @@ func (n notAllowedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	var body []byte
+
+	if r.Body != nil {
+		var err error
+		body, err = ioutil.ReadAll(r.Body)
+		if err != nil {
+			return nil, err
+		}
+		r.Body.Close()
```
wwwdata: this should be put 3 lines above and prefixed with a defer like in original `unmarshalRequest`.

api.go
```diff
@@ -185,12 +189,108 @@ func (n notAllowedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	} else {
+		data, err := unmarshalRequest(body, r, res.marshalers)
+		if err != nil {
+			return nil, errors.New("unmarshal_error: " + err.Error())
+		}
+		req.Data = data
+	}
+
+	if meta, ok := req.Data["meta"]; ok {
+		req.Meta = meta.(map[string]interface{})
```
wwwdata: Hm, does the unmarshaller already extract meta data? Or is this just a preparation when it's supported. I currently don't remember exactly. But if it does we have a meta field in `data` and the same again in `req.Meta` now, right?

api.go
```diff
@@ -185,12 +189,108 @@ func (n notAllowedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		var output []byte
+		var status int
+		contentType := "application/vnd.api+json"
+		header := w.Header()
+
+		request, err := res.buildRequest(r, params)
+
+		if err != nil {
+			output = []byte(fmt.Sprintf("{\"errors\": [{\"title\": \"%v\"}]}", err))
```
wwwdata: can you not use `res.marshalError` here again? this manual json look weird

api.go
```diff
@@ -185,12 +189,108 @@ func (n notAllowedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+func (res *resource) getMiddleware(handler func(Request) response) func(http.ResponseWriter, *http.Request, httprouter.Params) {
+	return func(w http.ResponseWriter, r *http.Request, params httprouter.Params) {
+
+		var output []byte
+		var status int
+		contentType := "application/vnd.api+json"
+		header := w.Header()
+
+		request, err := res.buildRequest(r, params)
```
wwwdata: if you use the middleware, `buildRequest` would be executed twice?

api.go
```diff
@@ -253,148 +356,141 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
...
 }
 
-func (res *resource) handleIndex(w http.ResponseWriter, r *http.Request, info information) error {
+func (res *resource) handleIndex(r Request) response {
 	pagination := newPaginationQueryParams(r)
 	if pagination.isValid() {
 		source, ok := res.source.(PaginatedFindAll)
 		if !ok {
-			return NewHTTPError(nil, "Resource does not implement the PaginatedFindAll interface", http.StatusNotFound)
+			return response{
```
wwwdata: why not use NewHTTPError here?

api.go
```diff
@@ -253,148 +356,141 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
...
+				Status: http.StatusNotFound,
+				Error: NewHTTPError(nil, "Resource does not implement the PaginatedFindAll interface", http.StatusNotFound),
+			}
 		}
 
-		count, response, err := source.PaginatedFindAll(buildRequest(r))
+		count, resp, err := source.PaginatedFindAll(r)
 		if err != nil {
-			return err
+			return response{
```
wwwdata: this could be avoided by just returning err and let the outer `handleError` call in place

api.go
```diff
@@ -185,12 +189,108 @@ func (n notAllowedHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			var err error
+
+			if resp.Error != nil {
+				log.Printf("api error: %v\n", resp.Error)
+				output, contentType, status = res.marshalError(resp.Error, r)
+			} else if resp.Data != nil {
+				output, contentType, err = res.marshalData(resp.Data, r)
+				if err != nil {
+					status = 500
+					output = []byte(fmt.Sprintf("{\"errors\": [{\"title\": \"%v\"}]}", err))
```
wwwdata: don't like this here too

api.go
```diff
@@ -598,118 +699,129 @@ func (res *resource) handleUpdate(w http.ResponseWriter, r *http.Request, ps htt
 	if structType.Kind() == reflect.Ptr {
 		structType = structType.Elem()
 	}
-
 	err = jsonapi.UnmarshalInto(ctx, structType, &updatingObjs)
 	if err != nil {
-		return err
+		log.Printf("unmarshal failed: %v\n", err)
```
wwwdata: debug log that needs removal?

api_interfaces.go
```diff
@@ -28,6 +29,11 @@ type CRUD interface {
 	Update(obj interface{}, req Request) (Responder, error)
 }
 
+type CRUDHooks interface {
+	BeforeHandle(*Request) *Responder
```
wwwdata: please remove the `*` from Responder interface type

api_test.go
```diff
@@ -1237,14 +1239,6 @@ var _ = Describe("RestHandler", func() {
 				"data": []interface{}{post1JSON},
 			}))
 		})
-
-		It("Extracts multiple parameters correctly", func() {
-			req, err := http.NewRequest("GET", "/v0/posts?sort=title,date", nil)
```
wwwdata: hey, we need that. it was very convencient to have these sort query parameters automatically extracted

error.go
```diff
@@ -56,35 +55,38 @@ type ErrorSource struct {
...
 }
 
 //marshalHTTPError marshals an internal httpError
-func marshalHTTPError(input HTTPError, marshaler ContentMarshaler) string {
+func marshalHTTPError(input HTTPError, marshaler ContentMarshaler) []byte {
 	if len(input.Errors) == 0 {
-		input.Errors = []Error{Error{Title: input.msg, Status: strconv.Itoa(input.status)}}
+		code := ""
+		if input.err != nil {
+			code = input.err.Error()
```
wwwdata: so if there is just 1 error the whole error is taken as the `code` field? Should the code not be something explicitly set and in this case, not be used? Becaues it's not explicitly set?

sharpner:
I appreciate the work you've done, but we can't merge it like this. 

I don't really get the intention behind this PR. You've refactored so much that I can't quite figure out what the goal of this is. It would be much easier if each change would be one single commit, or even PR. 

Additionally refactoring should be separated from feature changes. If they are done at the same time, we cannot make sure that nothing broke.

wwwdata:
Now that I am through:

- **getMiddleware** *you introduced this for the hooks and to clean stuff up, right?*
- **BeforeHandle and AfterHandle** *it is already possible to use before and after middlewares but you introduced this to only do it for a specific resource without parsing urls in a custom middleware again?*
- **Error code in response** *is not needed, because if there is an error, a client just returns an error in his resource method. If needed he can return our custom error type which already includes an error code*
- **meta data in request** *this is not working because meta data is currently not being extracted by the unmarshaler, or am i wrong?*
- **query params type change** *yes, that is good*
- **updated tests** *an important test was deleted, instead of changed*

api.go
theduke: Will do.

api.go
theduke: Will do.

api.go
theduke: It works already, since the unmarshaler just returns the raw unmarshaled map[string]interface{}.
req.

I agree that meta should be removed from data when it exists.
Will do.

api.go
theduke: Can do.

api_test.go
theduke: The reason I removed is because buildRequest depends on the Resource now, because it calls unmarshalRequest() which needs res.marshalers.

I'll find another way to test it.

wwwdata: thx :)

wwwdata:
@theduke we merged #170 now which enables the usage of a middleware chain and we merged #173 and have a default implementation of the `Responder` interface. Just to let you know if you plan to work again on this PR. 

So if you are planning to work on this, what exactly do you want to implement?

sharpner:
@theduke - I am going to close this PR for now. If you like you can simply open a new PR with your refactored changes.

PR #170, opened 2015-09-07
Added net/context support.

### LINKED ISSUE #169: Support passing a Context to handlers
Many advanced routing frameworks wrap handlers to also contain a Context implementation. Some are based on this https://godoc.org/golang.org/x/net/context and others are custom rolled.

If the community is open to it, I can submit a PR which adds this functionality. It should allow people to pass in per connection db handles, global and request configuration, logging handlers, etc.

interlock:
* added net/context as part of API info
* passed this ctx along to handlers
* inserted ctx as part of the Request, backwards compatible.
* very minor update to example that shows you can provide a ctx

closes #169

### DIFF
--- a/api.go
+++ b/api.go
@@ -191,6 +191,18 @@ type resource struct {
 	marshalers   map[string]ContentMarshaler
 }
 
+// middlewareChain executes the middleeware chain setup
+func (api *API) middlewareChain(c *APIContext, w http.ResponseWriter, r *http.Request) {
+	for _, middleware := range api.middlewares {
+		middleware(c, w, r)
+	}
+}
+
+// allocateContext creates a context for the api.contextPool, saving allocations
+func (api *API) allocateContext() *APIContext {
+	return &APIContext{}
+}
+
 func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, marshalers map[string]ContentMarshaler) *resource {
 	resourceType := reflect.TypeOf(prototype)
 	if resourceType.Kind() != reflect.Struct && resourceType.Kind() != reflect.Ptr {
@@ -224,24 +236,40 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 	}
 
 	api.router.Handle("OPTIONS", api.prefix+name, func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		c := api.contextPool.Get().(*APIContext)
+		c.reset()
+		api.middlewareChain(c, w, r)
 		w.Header().Set("Allow", "GET,POST,PATCH,OPTIONS")
 		w.WriteHeader(http.StatusNoContent)
+		api.contextPool.Put(c)
 	})
 
 	api.router.Handle("OPTIONS", api.prefix+name+"/:id", func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		c := api.contextPool.Get().(*APIContext)
+		c.reset()
+		api.middlewareChain(c, w, r)
 		w.Header().Set("Allow", "GET,PATCH,DELETE,OPTIONS")
 		w.WriteHeader(http.StatusNoContent)
+		api.contextPool.Put(c)
 	})
 
 	api.router.GET(api.prefix+name, func(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+		c := api.contextPool.Get().(*APIContext)
+		c.reset()
+		api.middlewareChain(c, w, r)
 		err := res.handleIndex(w, r, api.info)
+		api.contextPool.Put(c)
 		if err != nil {
 			handleError(err, w, r, marshalers)
 		}
 	})
 
 	api.router.GET(api.prefix+name+"/:id", func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		c := api.contextPool.Get().(*APIContext)
+		c.reset()
+		api.middlewareChain(c, w, r)
 		err := res.handleRead(w, r, ps, api.info)
+		api.contextPool.Put(c)
 		if err != nil {
 			handleError(err, w, r, marshalers)
 		}
@@ -254,7 +282,11 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 		for _, relation := range relations {
 			api.router.GET(api.prefix+name+"/:id/relationships/"+relation.Name, func(relation jsonapi.Reference) httprouter.Handle {
 				return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+					c := api.contextPool.Get().(*APIContext)
+					c.reset()
+					api.middlewareChain(c, w, r)
 					err := res.handleReadRelation(w, r, ps, api.info, relation)
+					api.contextPool.Put(c)
 					if err != nil {
 						handleError(err, w, r, marshalers)
 					}
@@ -263,7 +295,11 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 
 			api.router.GET(api.prefix+name+"/:id/"+relation.Name, func(relation jsonapi.Reference) httprouter.Handle {
 				return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+					c := api.contextPool.Get().(*APIContext)
+					c.reset()
+					api.middlewareChain(c, w, r)
 					err := res.handleLinked(api, w, r, ps, relation, api.info)
+					api.contextPool.Put(c)
 					if err != nil {
 						handleError(err, w, r, marshalers)
 					}
@@ -272,7 +308,11 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 
 			api.router.PATCH(api.prefix+name+"/:id/relationships/"+relation.Name, func(relation jsonapi.Reference) httprouter.Handle {
 				return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+					c := api.contextPool.Get().(*APIContext)
+					c.reset()
+					api.middlewareChain(c, w, r)
 					err := res.handleReplaceRelation(w, r, ps, relation)
+					api.contextPool.Put(c)
 					if err != nil {
 						handleError(err, w, r, marshalers)
 					}
@@ -283,7 +323,11 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 				// generate additional routes to manipulate to-many relationships
 				api.router.POST(api.prefix+name+"/:id/relationships/"+relation.Name, func(relation jsonapi.Reference) httprouter.Handle {
 					return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+						c := api.contextPool.Get().(*APIContext)
+						c.reset()
+						api.middlewareChain(c, w, r)
 						err := res.handleAddToManyRelation(w, r, ps, relation)
+						api.contextPool.Put(c)
 						if err != nil {
 							handleError(err, w, r, marshalers)
 						}
@@ -292,7 +336,11 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 
 				api.router.DELETE(api.prefix+name+"/:id/relationships/"+relation.Name, func(relation jsonapi.Reference) httprouter.Handle {
 					return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+						c := api.contextPool.Get().(*APIContext)
+						c.reset()
+						api.middlewareChain(c, w, r)
 						err := res.handleDeleteToManyRelation(w, r, ps, relation)
+						api.contextPool.Put(c)
 						if err != nil {
 							handleError(err, w, r, marshalers)
 						}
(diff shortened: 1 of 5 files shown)


interlock:
Not entirely happy about pushing the ctx into the Request, nor that there is no great way to insert middle wear modified ctx's.

I want to look at how the router works a bit more, to see if we can get a per request ctx made that flows through. This may require us to break CRUD and related interfaces to insert a ctx more cleanly.

Alternately, we could create structs to represent the calling to the CRUD and relationship implementations, which would allow us to add request/response complexity in the future without breaking the interface prototypes. See the AWS golang implementation for a somewhat overwhelming, but good example of how that would work: http://docs.aws.amazon.com/sdk-for-go/api/service/sqs.html

Comments, suggestions and hi-fives welcome.

wwwdata:
hm, now every request would have the same context that was created when initializing api2go. This is not the intended behavior right? Every incoming request should generate a context object which is then passed through the middleware chain which is in our case, not existing because we just use httprouter and the request ends there. 

so if you have a middleware chain with an already initialized request chain and api2go is at the end of this chain, it would then be useful to access the context struct?

interlock:
You nailed it @wwwdata . I am looking at how httprouter works right now to see how practical it would be to get some middleware going with a ctx. At the very least, we could probably create a func prototype and method on API to insert middlewares if httprouter can not handle it.

sharpner:
If it helps to provide a cleaner implementation, we currently still allow breaking changes if necessary.

interlock:
Alright @sharpner 

I am going to adjust this branch to be Context and Middleware.

I'll get a middleware inserted first. I want to echo what @wwwdata pointed out, that having a ctx per request is ideal. You can pull a sql connection from a pool that handles that request. Assign a request ID and use it in subsequent logging and so on. Using the net/context also has some advanced features for forcing deadlines and handling async errors. We use this in another service we run and it sets you up to create rock solid 12 factor services.

I think we will have to break anything implementing an interface that returns `Responder`. In all those cases, we are going to want to pass along the context. Internally there will obviously be a lot of refactoring to deal with those context's being passed around.

I am going to leave the idea of doing the aws-go for now, but I do believe it is a great way to setup functions for testing and exposing new information without breaking the public interfaces. `information` currently attempts to fill this role, but no reason we couldn't make a struct specific to each CRUD/Relation to encapsulate these.

interlock:
This is passing all but one test. Here is an overview of what I changed:

Instead of reinventing an implementation of middleware and context passing, I found `gin` is built on the same httprouter. While I am not a fan of the "one context to rule them all", it works and is something I can live with. On the upside, gin.Context implements net/context so anyone who wants to cast the down and use more generic handlers is free to do so.

You now have access to the gin.Engine where middleware can be added:

`api.Router().Use(...)` Handler functions for gin confirm to `func handle(c *gin.Context)`. Request and Writer are both nested in the uber gin.Context as `Request` and `Writer`. Gin provides plenty of utility functions around handling requests.

CRUD and related interfaces now expect (*gin.Context, ...) as a first param. The convention set out for using net/context is it comes as the first parameter, which is reflected in the refactor through out `api2go`.

There are some context related functions to pull out a reference to the `API` and the special split QueryParams used throughout. See `context.go` for those.

### Next up

* I think it makes sense to make an `api2go.APIContext` that embeds the `gin.Context`. While I do not see any need to add anything more to that context. It does allow us to change our minds in the future without breaking everyones implementations.
* I am going to suggest another fairly massive breaking change to change our CRUD interfaces to take a InputStructs and OutputStructs. It would be another breaking change built on top of what this PR is. I will open an issue to propose this and discuss.
* Solve that last failing, but relatively benign test that is failing.

wwwdata:
So basically everything is now build ontop of the gin-gonic framework. We initially just wanted to use a context, which is now possible because it can be injected into the gin-gonic router which has a [Use](https://godoc.org/github.com/gin-gonic/gin#RouterGroup.Use) method that accepts their custom middleware type which takes their [context](https://godoc.org/github.com/gin-gonic/gin#Context)

This now forces a hard vendor lock of the gin-gonic framework. We initially just wanted to be standard compatible, which means only use the normal golang [http.Handler](https://godoc.org/net/http#Handler). The gin context does a lot of things like handling the router params, the request and the response and also supports fancy stuff like JSON encoders. I don't know this is a big overkill and contradicts our initial intention to just be very minimal. 

I understand that a context is important if you build your web application that has a lot of custom middlewares that need this to pass all kinds of stuff in between them and to also set Deadlines when a request has to time out etc...
But if you want to do this, is it not the better choice to just use our marshaler/unmarshaler and build the rest of the complex framework with gin-gonic by yourself? I just have all the standard users in mind that just need normal a `http.Handler` and use the standard golang middlewares to chain authentication before our routing. 

All this middleware stuff is more interesting if you are in total control of the generated routes, like in this [example from gin](https://github.com/gin-gonic/gin#using-basicauth-middleware). Where they are protecting a sub-tree `admin` of their routes with a basic auth middleware. Creating the routes by yourself gives you fine grained control of how everything is put together.

So my question is: Is this worth it? Use all the gin stuff just to be able to chain one middleware before everything which uses their context, that get's passed to one more method inside the api2go resource and no more endpoints before that? We pre-generate all the routes and once they are generated, they cannot be accessed or manipulated from the outside anymore...

interlock:
Good points @wwwdata 

What is not obvious here about context/middlewares as presented:
* Where do you insert validating authentication and associating identity. A middleware  is the most common place to do this.
* If you are accessing the db, you are probably using a connection pool. Middleware can pass the pool along to the CRUD implementations OR assign a connect by request context.
* Rate limiting.
* CORS, if you need it.
* Gzip, though I think it's on by default in httprouter.
* Edge Caching.
* Logging/Auditing.
* You can actually insert the gin.Engine into a regular http.Handler chain, use any framework you like with the generated routes from api2go (this is how we are currently testing it actually).

If we carry on with this PR, we can submit some updates to the example project that make it more clear how middleware would be used (moving the storage drivers in to the ctx, simple auth, etc).

I am not a fan of vendor lock in either, but httprouter was a form of vendor lock already. It required handlers to have a third param: `httprouter.Params`. 

In our test implementation of api2go our team identified two areas that limited our ability to execute quickly. Passing along auth/ident information and working with our DB/Cache pools. The implementation gin provides is concurrency safe which helps us get the most out of of `golang`. The alternative for api2go is to use the registry pattern, which if you make it concurrency safe becomes a deadlocking and blocking risk (been there, not fun).

Alternatively, we duplicate the functionality of their middleware and context handling. Make that fit the existing one of httprouter handler pattern. I favoured using gin because they are built on httprouter (fastest of the routers).

You can also unwrap gin.Context to call the usual http.Handler type handlers in a closure/generator.

interlock:
Removed the WIP on the PR, all lights are green and it appears to work.

wwwdata:
no, httprouter was no vendor lock in, because it was only internally used and the user could use standard http handlers because it was completely transparent. Gin on the other hand is a vendor lock in now because a `gin.Context` is now in all the methods of api2go resources. 

one of the main problems now is, that the `gin.Context` allows the user of api2go to directly write into the response, right? But we have our `Responder` and `error` return values which should be used for that. I appreciate your work on that and I really think because of the good reasons you mentioned earlier that a context is a very good thing to have but the `gin.Context` is too much for us and can be abused in a lot of cases. in addition to that, the custom `Request` object was really useful because we had some already parsed query parameters in it for example. And by keeping a custom `Request` struct, we can always extend it without breaking the api. Why not just put a context in there which does not allow to manipulate the http resonse in any way but which can be used to pass through db connections or authentication information and so on? In the simplest form I guess a `map[string]interface{}` could do that

I am imagining a usage example of a db middleware like this:

```go
// Request type will be used but PlainRequest gets replaced with the new context. 
// Only the *http.Request is directly accessible, the rest is managed via Get/Set Methods and so on
type Context struct {
	Request        *http.Request
	responseWriter http.ResponseWriter
	values         map[string]interface{}
}

type Request struct {
	Context      api2go.Context
	QueryParams  map[string][]string
	Header       http.Header
}


func dbMiddleware(context api2go.Context) {
	// get your db with connection pool etc...
	myDBStruct := ...
	context.Set("db", myDBStruct)
	
	// if we do this, context would have to manage request/reponse internally like in gin too,
	// but we will not expose that to the user
	context.Next()
}

api := api2go.NewAPI("/v1")
api.UseMiddleware(dbMiddleware)

// middleware would be used in order, to dbMiddleware always before someOtherMiddleware and so on
api.UseMiddleware(someOtherMiddleware)

http.ListenAndServe(":31415", api.Handler())
```

What does @sharpner think?

sharpner:
Yeah I agree with @wwwdata,  although it would be perfect if we could use an interface
```go
type Contextor interface {
  Set(string, interface{})
  Next()
}
```
that way the user could decide wether to use gin.Context or the Google one, or a custom one.

interlock:
Thanks @sharpner and @wwwdata for the feedback. Nothing out of line here, lets  iterate!

What I am hearing here are these concerns:

_We do not what to pass anything more than what is required to the CRUD interfaces, where did Request and QueryParams go?_

This was actually going to be my next recommendation. I agree that if we build and pass `Request` we future proof the API to some extend. So, I propose we follow the pattern in this library http://docs.aws.amazon.com/sdk-for-go/api/service/sqs.html

```go
type FindOneInput struct {
ID string
QueryParams Params
Request http.Request
Context context.Context
}
```

```go
type FindOneOutput struct {
meta map[string]interface{}
result interface{}
code int
}
```

This allows us to add more information to a `FindOne` implementation down the line without breaking the interface. It limits the surface area we expose as well, just what you should need to do the action.

As for QueryParams, it was in the context and accessible via ContextQueryParams(...). If we create Input/Output structs this is moot, but handy play to put it just before we build the Input struct.

_Middleware_

Sounds like we still want it, but we don't want the mega context gin comes with nor the handler func they implement. While httprouter does require a custom handler like gin https://godoc.org/github.com/julienschmidt/httprouter#Handle it closer resembles the standard handler http://golang.org/pkg/net/http/#HandlerFunc. Combining our need for context, the proposed handler would look like this:

`type func APIHandler(*context.Context, ResponseWriter, *Request, httphandler.Params)`

_Context is still good, but lets stick to the standard DRY net/context_

I will refactor to use a standard net/context, light and easy. This is actually quite easy, as how we used Context in a few spots assumed the basic context.

wwwdata:
sounds good for me! 

I don't know if we need the `httphandler.Params` in the `APIHandler` signature. I think we can leave them out since it's only used internally. Or do you have a use case where a middleware also needs the params that our internal generated routes are using?

interlock:
@wwwdata we could put it in the context if someone needs it.

I'll pull back the gin changes and rejigger everything. Hopefully have am updated PR ready for tomorrow.

wwwdata:
awesome, thank you very much!

sharpner:
yeah! thx

interlock:
Very basic middleware support with context.

Next up refactor the CRUD to use Input/Output structs.

context.go
```diff
@@ -0,0 +1,69 @@
+package api2go
+
+import (
+	"golang.org/x/net/context"
+	"time"
+)
+
+// APIContext api2go context for handlers, nil implementations related to Deadline and Done.
+type APIContext struct {
```
sharpner: If I haven't missed anything, apiContext could be private and use an public Interface right? Since the user only implements a middleware with it being the first parameter.

wwwdata: an interface for that could be added and used as type everywhere, in that case the enduser can use this APIContext or implement his own if he wants too.

api.go
```diff
@@ -224,24 +236,40 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD, ma
 	}
 
 	api.router.Handle("OPTIONS", api.prefix+name, func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		c := api.contextPool.Get().(*APIContext)
```
wwwdata: I really like the usage of a sync.Pool. Never done that before but I looked into the gin code and they use it too. It makes sense because otherwise a APIContext struct would have to be allocated for every request and then garbage collected. In that way, it can be used again without new allocation/garbage collection, right?

When everything is working in api2go we should look for more performance optimisation stuff :)

sharpner: Yeah awesome commit thus far.

wwwdata:
@interlock what exactly do you have in mind for the CRUD refactoring? Can you explain? This commit looks like it could already be merged and we would have the context feature.

context.go
interlock: The Context interface leaves a lot to the implementer to expose how you set a value on it. Which means the `Value` func has to match implementation. Unfortunately, that means if we want to expose people setting values in context from middlewares we need to provide that as a public struct.

interlock:
@wwwdata We need to sneak the context into the CRUD then, possibly in the buildRequest function?

If we go that far, I say merge it at that point and we will make a PR that shows what we mean with the CRUD refactor.

context.go
wwwdata: I think @sharpner means that the api2go `APIContext` should have an additional interface, so that a user can extend/change the `APIContext` easily. And yes, it must be public in order to be jused by somebody that does not want to implement it by himself. But all our internal methods would require the interface. That makes sense or?

wwwdata:
@interlock yes buildRequest should take the context and put it in our Request struct.

context.go
sharpner: exactly! ヾ(◠‿◠ ✿)
The signature should start like this ```middlewareChain(c api2go.ContextInterfaceThingy,...)``` where APIContext implements this interface with its methods...

interlock: Ah, I got ya. That's an easy one then ;-)

That would imply we want to allow people to provide a New func to the pool then if they choose to implement the interface, so I'll add that at the same time.

sharpner: Yeah, forgot to mention that! perfect

interlock:
Update to reflect context as an interface and building `Request` with our context in it.

wwwdata:
cool, what is that merge commit at the end? are you sure that you do not have some commits that are not wanted now? you should rebase your stuff onto our latest master branch

interlock:
@wwwdata That's me screwing up a rebase actually. Sorry about that :-( Pre-coffee merges are the worst

interlock:
@wwwdata lets see if that fixes it up... always feel terrible doing a `push --force`

wwwdata:
yes, commits are looking good new. It's totally fine to do force pushes in pull requests.. When I am working i rebase and squash my commits a lot to clean them up. It is always a good thing to have commits that can be easily understood and are separated into logical steps that make sense. 

I will check all changes again later. This is complete now, right? So if we are good with it, we can merge?

interlock:
Yup, this is feature complete now.

The example was not updated, but it may make things more confusing as a simple example if we did that. May require some doc examples at some point?

wwwdata:
Yes you could update the readme of you like. But I can do that too later.

wwwdata:
nice work, going to merge this now :+1:

PR #172, opened 2015-09-10
use `jsonapi:"-"` tag to ignore fields, fixes #171

### LINKED ISSUE #171: jsonapi internal ignore tag
When working with different database drivers, that rely on json themselves, it is possible that I need to deserialize fields, that I do not want to serialize via api2go. In that case, it was necessary to use the `json:"-"` ignore tag. This is a problem because i want these fields to be serialized/deserialized when talking to my database via json. 

Therefore, it would be a nice feature to have a `jsonapi:"-"` ignore tag, that only ignores if serialized via our jsonapi marshaler.

Of course it is always a possibility to use a different struct to talk to the database, but it's unnecessary overhead if it is not needed and everything can be achieved with one struct.

wwwdata:
also fixed a model in the examples project which forgot to ignore the ID
field. Tests were wrong too and expected an id field inside attributes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,11 @@ comments that belong with a has-many relation to the post.
 
 ```go
 type Post struct {
-  ID          int
+  ID          int       `jsonapi:"-"` // Ignore ID field because the ID is fetched via the
+                                      // GetID() method and must not be inside the attributes object.
   Title       string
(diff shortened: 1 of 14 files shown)


sharpner:
nice! Should we add one additional test that has both ```json:"-"``` and ```jsonapi:"-"``` ignore tags?

PR #173, opened 2015-09-11
api2go.Response as default for api2go.Responder

sharpner:
This can be useful for very basic applications that don't require a special `api2go.Responder` implementation


wwwdata:
:+1:

PR #180, opened 2015-09-24
[WIP] Two new interfaces for baseurl resolving

sharpner:
this commit allows per request dynamic handling of baseUrls
it also deprecates the old NewAPIWithMarshalers in favor
of the new NewAPIWithMarshalling (name is subject to change).

It is fully backwards compatible.

### DIFF
--- a/examples/resolver/resolver.go
+++ b/examples/resolver/resolver.go
@@ -1,0 +1,28 @@
+package resolver
+
+import (
+	"fmt"
+	"net/http"
+)
+
+//RequestURL simply returns
+//the request url from REQUEST_URI header
+//this should not be done in production applications
+type RequestURL struct {
+	r    http.Request
+	Port int
+}
+
+//SetRequest to implement `RequestAwareResolverInterface`
+func (m *RequestURL) SetRequest(r http.Request) {
+	m.r = r
+}
+
+//GetBaseURL implements `URLResolver` interface
+func (m RequestURL) GetBaseURL() string {
+	if uri := m.r.Header.Get("REQUEST_URI"); uri != "" {
+		return uri
+	}
+
+	return fmt.Sprintf("https://localhost:%d", m.Port)
+}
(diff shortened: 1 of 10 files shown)


examples/resolver/resolver.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+	"net/http"
+)
+
+//RequestURL simply returns
+//the request url from REQUEST_URI header
+//this should not be done in production applications
+type RequestURL struct {
```
wwwdata: cool, but we should provide one or two resolvers that can be used out of the box in my opinon. Just to get ppl started quickly without implementing a lot

sharpner: In order to get started you don't need one. We could not deprecate the baseURL constructor method, but we would than need a better name for `NewAPIWithMarshalling`

sharpner: I wouldn't know how to implement a generic dynamic Resolver. Any ideas?

sharpner: I could do a CallbackResolver so that people would only need to implement a callback function instead of a struct with two methods, but that's about it :/

wwwdata: Hm, I guess it can be done like in your examples, to just take the base url from the request. Or take the base url out of another provided header that is configurable like `X-Forwarded-For` or sth.

wwwdata: that callback idea is also nice. You get in the request, and output the base url, that would be the function signature, right?

sharpner: But it would be a little more complex than that. It should have a whitelist of allowed TLDs and maybe some regex handling .. don't you think?

sharpner: it would be somewhat like this ```go NewCallbackResolver(func(r http.Request) string {})```

wwwdata: I don't know, it's not that the base url is fake-able if there is a proxy like nginx before it or?

sharpner: No, then you're safe. But not everyone does it that way.

PR #182, opened 2015-09-25
[RFC] Extract router to be an interface

sharpner:
this is particulary useful since more and more people
are trying to use api2go with gin router.

Since we have built so many new features I will restructure our constructors within this pr.
This will probably deprecate every constructor so far except the simple `NewAPI(namespace string)` one.

### DIFF
--- a/api_public.go
+++ b/api_public.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/manyminds/api2go/jsonapi"
 	"github.com/manyminds/api2go/routing"
 )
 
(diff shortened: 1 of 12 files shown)


sharpner:
@wwwdata I think I'm done. Not sure about the naming though. I hate having "routing" and "adapter" packages. Any ideas?

I tried implementing the gin adapter in its own package, so that you won't need the whole gin framework to use api2go.

sharpner:
So I moved the adapter over to https://github.com/manyminds/api2go-adapter - it works but the build will fail until this branch got merged ;-) 

You can copy paste the example in the readme to check it out.

sharpner:
What do you think, should we move the http adapter to api2go-adapter as well? 
Even if it's the default one, It could be profitable treat him equally to the others.

wwwdata:
I think we should keep the default in this package.

sharpner:
kk. Anything else?

api_public.go
```diff
@@ -27,57 +30,35 @@ type API struct {
...
+//Router returns the specified router on an api instance
+func (api API) Router() routing.Routeable {
 	return api.router
 }
 
-// AddResource registers a data source for the given resource
-// At least the CRUD interface must be implemented, all the other interfaces are optional.
-// `resource` should be either an empty struct instance such as `Post{}` or a pointer to
-// a struct such as `&Post{}`. The same type will be used for constructing new elements.
-func (api *API) AddResource(prototype jsonapi.MarshalIdentifier, source CRUD) {
```
wwwdata: hm is there any reason you moved these 2 methods to the other file? They are public so why not let them be in `api_public.go`?

wwwdata:
and apart from that you prefer the comment style without whitespace after `//`. I always did a whitespace.. I think we should also do that in the same way, or? Too bad the go linter has no opinion about that.

api_public.go
sharpner: no idea lol. Ill change that.

sharpner:
:D I'll change it. Yeah It can't get used to typing a whitespace at the beginning of a line :)

wwwdata:
ok you're done and we can merge right?

PR #188, opened 2015-10-02
marshaling support for embedded structs

### LINKED ISSUE #187: Marshaling of struct compositions fails
When I compose a struct and then marshal it using the normal JSON encoder, all fields including those inherited from the anonymous field will be direct child objects of the composed struct. Using `jsonapi` however it creates an extra node for the anonymous field and fails to generate valid JSON.

Please see the [`marshal_composition_test.go`](https://gist.github.com/christianklotz/1d01b08027d7866d49f4) and the example given below for more.

Is this intended behaviour? If so, what would be your suggestion to deal with struct compositions?

### Example
Encoding a `TaggedPost` which is composed of a `SimplePost` and an additional field `Tag` using the  default JSON encoder gives you the following result, also see http://play.golang.org/p/JobhDfJOgu for the playground.

#### Default JSON
```json
{
  "ID":"first",
  "Title":"First Post",
  "Text":"Lipsum",
  "Size":0,
  "Created":"2014-11-10T16:30:48.823Z",
  "Tag":"important"
}
```

#### JSONAPI – Expected Result

```json
{
  "data": {
    "type": "taggedPosts",
    "id": "first",
    "attributes": {
      "title": "First Post",
      "text": "Lipsum",
      "size": 0,
      "create-date": "2014-11-10T16:30:48.823Z",
      "tag": "important"
    }
  }
}
```

#### JSONAPI – Actual Result

```json
{
  "data": {
    "type": "taggedPosts",
    "attributes": {
      "tag": "important",
      "simplePost": {
        ID: "first",
        Title: "First Post",
        Text: "Lipsum",
        Internal: "",
        Size: 0,
        Created: {
          sec: 63551233848,
          nsec: 823000000,
          loc: {
            name: "UTC", 
            zone: nil, 
            tx: nil, 
            cacheStart: 0, 
            cacheEnd: 0, 
            cacheZone: nil
          },
        },
      },
    },
    "id": "first",
  }
}
```

wwwdata:
this will resolve #187 

todo:
- [x] unmarshaling

### DIFF
--- a/jsonapi/marshal.go
+++ b/jsonapi/marshal.go
@@ -388,9 +388,8 @@ func getStructFields(data MarshalIdentifier) map[string]interface{} {
 		}
 
 		field := val.Field(i)
-		keyName := Jsonify(valType.Field(i).Name)
 
-		//skip private fields
+		// skip private fields
 		if !field.CanInterface() {
 			continue
 		}
@@ -400,6 +399,19 @@ func getStructFields(data MarshalIdentifier) map[string]interface{} {
 				continue
 			}
 		}
+
+		// check for embedded structs and also extract all fields of them into result
+		if embeddedStruct, ok := field.Interface().(MarshalIdentifier); ok {
+			embeddedFields := getStructFields(embeddedStruct)
+			for k, v := range embeddedFields {
+				result[k] = v
+			}
+
+			// skip embedded struct itself
+			continue
+		}
+
+		keyName := Jsonify(valType.Field(i).Name)
 
 		name := GetTagValueByName(valType.Field(i), "name")
 		if name != "" {
--- a/jsonapi/unmarshal.go
+++ b/jsonapi/unmarshal.go
@@ -258,15 +258,7 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type, 
 					fieldType, found := val.Type().FieldByName(fieldName)
 					if !found {
 						//check if there is any field tag with the given name available
-						for x := 0; x < val.NumField(); x++ {
-							tfield := val.Type().Field(x)
-							name := GetTagValueByName(tfield, "name")
-							if tfield.Tag.Get("jsonapi") != "-" && strings.ToLower(name) == strings.ToLower(fieldName) {
-								field = val.Field(x)
-								found = true
-							}
-						}
-
+						field, found = getFieldByTagName(val, fieldName)
 					}
 					if !found || fieldType.Tag.Get("jsonapi") == "-" {
 						return fmt.Errorf("invalid key \"%s\" in json. Cannot be assigned to target struct \"%s\"", key, targetStructType.Name())
@@ -318,6 +310,30 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type, 
 	}
 
 	return nil
+}
+
+// check if there is any field tag with the given name available
+func getFieldByTagName(val reflect.Value, fieldName string) (field reflect.Value, found bool) {
+	for x := 0; x < val.NumField(); x++ {
+		// check if there is an embedded struct which needs to be searched
+		if val.Field(x).CanAddr() && val.Field(x).Addr().CanInterface() {
+			_, isEmbedded := val.Field(x).Addr().Interface().(UnmarshalIdentifier)
+			if isEmbedded {
+				field, found = getFieldByTagName(val.Field(x), fieldName)
+				return
+			}
+		}
+
+		// try to find the field
+		tfield := val.Type().Field(x)
+		name := GetTagValueByName(tfield, "name")
+		if tfield.Tag.Get("jsonapi") != "-" && strings.ToLower(name) == strings.ToLower(fieldName) {
+			field = val.Field(x)
+			found = true
+		}
+	}
+
+	return
 }
 
 // setFieldValue in a json object, there is only the number type, which defaults to float64. This method convertes float64 to the value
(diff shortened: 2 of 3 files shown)


jsonapi/marshal.go
```diff
@@ -387,9 +387,22 @@ func getStructFields(data MarshalIdentifier) map[string]interface{} {
...
+				}
+
+				// skip embedded struct itself
+				continue
+			}
+		}
 		keyName := Jsonify(valType.Field(i).Name)
 
-		//skip private fields
+		// skip private fields
```
sharpner: Can we do this check a little bit earlier? ATM you're doing it twice.

wwwdata: Aah yes I overlooked that. We just pull that up and the code is better.

wwwdata:
@sharpner @christianklotz Unmarshaling now works too. Can you please review?

sharpner:
retest this please

jsonapi/unmarshal.go
```diff
@@ -320,6 +312,30 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type,
 	return nil
 }
 
+// check if there is any field tag with the given name available
+func getFieldByTagName(val reflect.Value, fieldName string) (field reflect.Value, found bool) {
+	for x := 0; x < val.NumField(); x++ {
+		// check if there is an embedded struct which needs to be searched
+		if val.Field(x).CanInterface() && val.Field(x).CanAddr() {
+			_, isEmbedded := val.Field(x).Addr().Interface().(UnmarshalIdentifier)
```
sharpner: shouldn't you check if val.Field(x).Addr().CanInterface() as well?

wwwdata: you are probably right, i will just change this, we only need to know if it `CanAddr()` and `CanAddr().CanInterface()`

sharpner:
retest this please

sharpner:
otherwise I think it's ok! Hurts to see that unmarshal grows even more complex but nothing we can do about it :P

christianklotz:
I just found out one issue with this implementation. It only works if the `GetID` method is implemented in the anonymous field type, so in our previous example `SimplePost` implemented it.

### Correct Result for inherited `GetID`
```go
type Foo struct {
	Name string `jsonapi:"name=name"`
}

func (f Foo) GetID() string {
	return "1"
}

type FooBar struct {
	Foo
	Age int `jsonapi:"name=age"`
}

func main() {
	fb := FooBar{Foo{"John"}, 12}
	s, _ := jsonapi.MarshalToJSON(fb)

	fmt.Println(string(s))
}
```

```json
{
  "data": {
    "attributes": {
      "age": 12,
      "name": "John"
    },
    "id": "1",
    "type":"fooBars"
  }
}
```
### Wrong Result for directly implemented `GetID`

```go
type Foo struct {
	Name string `jsonapi:"name=name"`
}

type FooBar struct {
	Foo
	Age int `jsonapi:"name=age"`
}

func (f FooBar) GetID() string {
	return "1"
}

func main() {
	fb := FooBar{Foo{"John"}, 12}
	s, _ := jsonapi.MarshalToJSON(fb)

	fmt.Println(string(s))
}
```

```json
{
  "data": {
    "attributes": {
      "age": 12,
      "name": {
        "Name": "John"
      }
    },
    "id": "1",
    "type":"fooBars"
  }
}
```

wwwdata:
the `GetID` depends. If you define another `GetID` Method on the struct that embeds another struct, the outer most `GetID` will get called. That is how this syntactic sugar of go works, as far as I understood it. But the problem here is the weird issue with the `name` field, right? This is interesting, I will add a test for this and find out the problem

christianklotz:
Just saw there was a small typo, it's corrected now it does encode it as `"foo":{"Name":"John"}` instead of `"name":"John"`. So the problem is that it's not resolving the fields of the anonymous field as part of the outer struct unless the `GetID` method is defined on the anonymous field's type.

wwwdata:
that is the intended behavior! I checked if the inner struct can be marshalled by checking if it implements the `MarshalIdentifier` interface. This does make sense or? You will use this when you already have an api2go struct but just want to make a larger version of it with some extra fields quickly.

wwwdata:
or do you have another idea? Maybe explain to us what exactly you want to do with this embedding :smile:

christianklotz:
Ah, ok – great. I guess it does make sense if you look at it like that. The scenario I was coming from was my custom struct embeds a more top level struct as shown in the example below. In this case the `GetID` method is defined on the outer struct.

```go

type Customer struct {
	Name string
}

type User struct {
	ID int
	Customer
}

func (u User)GetID() string {
	return strconv.Itoa(u.ID)
}
```

wwwdata:
hm, so you do not want to have api routes for the `Customer` struct? It's like an `abstract base class` if you describe it in object oriented terms, right? If we want to do this, we have to thing about something new, like another interface that this abstract base struct is implementing to signal to the marshaler, that all it's fields need to be pulled out. But I don't know if that is such a good idea? what does @sharpner think?

christianklotz:
Yeah, pretty much. I have a business logic layer that doesn't care much
about the API routes, handlers or even the database. The web service layer
then implements its own structs which embed the business logic structs and
adds whatever fields necessary.

It's already a huge improvement though :)

wwwdata:
another approach for you could be to have mappers that map the business structs into a single api2go compatible struct that has all the necessary things like field renaming, interface implementations and so on. This would give you even more flexibility and you just have one place to configure json related things. With the embed solution, in order to rename or exclude fields, you have to insert the json tags in 2 places.

christianklotz:
Sounds good.

Issue #190, opened 2015-10-05
Unmarshal to *time.Time fails


mantenie:
We use in our case a struct with pointer type ```*time.Time```. Marshaling to json works fine. But ```PATCH```or ```POST```of a struct with a datetime-string fails with an error like ```Could not set field 'Field-name'. Value '2015-10-04T08:54:47.569Z' had wrong type```.

wwwdata:
can you provide a short failing test?

wwwdata:
just for the record, if somebody finds this ticket, it is discouraged to use a pointer to `time.Time` in a struct, so we now skip the time field if it has zero value, see: https://golang.org/pkg/time/#Time

wwwdata:
closing this because #192 was merged in.

PR #191, opened 2015-10-05
Make type of *time.Time possible as property value

mantenie:
This fixes issue https://github.com/manyminds/api2go/issues/190.

### DIFF
--- a/jsonapi/unmarshal.go
+++ b/jsonapi/unmarshal.go
@@ -285,6 +285,13 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type, 
 							}
 
 							field.Set(reflect.ValueOf(t))
+						case *time.Time:
+							t, err := time.Parse(time.RFC3339, plainValue.String())
+							if err != nil {
+								return errors.New("expected RFC3339 time string, got '" + plainValue.String() + "'")
+							}
+
+							field.Set(reflect.ValueOf(&t))
 						default:
 							if field.CanAddr() {
 								switch field.Addr().Interface().(type) {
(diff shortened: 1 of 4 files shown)


jsonapi/unmarshal.go
```diff
@@ -285,6 +285,13 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type,
 							}
 
 							field.Set(reflect.ValueOf(t))
+						case *time.Time:
```
wwwdata: ah yes this makes sense, `time.Time` was also treated as a special case so it must be the same with the pointer type. We really have to start with refactoring #111

wwwdata:
hm, the go documentation says that you should not store `time.Time` as a pointer inside a struct see

https://golang.org/pkg/time/#Time

A Time represents an instant in time with nanosecond precision.

> Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time. A Time value can be used by multiple goroutines simultaneously.

> Time instants can be compared using the Before, After, and Equal methods. The Sub method subtracts two instants, producing a Duration. The Add method adds a Time and a Duration, producing a Time.

> The zero value of type Time is January 1, year 1, 00:00:00.000000000 UTC. As this time is unlikely to come up in practice, the IsZero method gives a simple way of detecting a time that has not been initialized explicitly.

so yes, we could merge this, but this problem can be solved by just using `isZero` from time package

wwwdata:
what was the problem that made you use `*time.Time`? Was ist that zero values like dates from 1970 should be set to null, or be ignored in the json? I think the correct solution would be to omit these zero value dates from the json

wwwdata:
closing this now because of #192

PR #192, opened 2015-10-06
Marshal an empty datetime to zero time.

mantenie:
This is a second approach to solve the problem, that a time value can be optional.
See: https://github.com/manyminds/api2go/pull/191 and https://github.com/manyminds/api2go/issues/190


wwwdata:
ok, so if there is a zero valued time.Time field now, it get's automatically omitted from the json. I think this is ok, @sharpner does also approve?

Issue #195, opened 2015-10-09
jsonapi handling errors?


faeronsayn:
How can we marshal errors into jsonapi compatible errors?

wwwdata:
Hi, if you are using our api functionality, just return an instance of our `HTTPError` as `error`. This will get marshalled correctly: https://godoc.org/github.com/manyminds/api2go#HTTPError
If you return any other normal go like error, then this will also be marshalled into a json api compatible error, but always with the default status code 500. 

You can create a new error with `NewHTTPError` https://godoc.org/github.com/manyminds/api2go#NewHTTPError

@faeronsayn do you think we should add this to our documentation in this chapter? https://github.com/manyminds/api2go#building-a-rest-api

faeronsayn:
Should definitely be in the documentation. But I was wondering how we can do it without using the API functionality.

I'm simply using the `jsonapi` package and it'd be nice if there was a way to do it with just the package.

wwwdata:
ok in that case, you can either do it yourself, because you are only using our marshal/unmarshal functionality or you can use the structs from the api package. It's your choice. I will add an example for our api users in the documentation.

wwwdata:
our api is very flexible. We now support adapters which allows you to use our generated routes with every framework that you like. We already wrote an adapter for the popular gin framework: https://github.com/manyminds/api2go#using-api2go-with-the-gin-framework

So, if you are planning to do a normal rest api with the standard routes, this is the easy way to go.

faeronsayn:
thanks :+1:

PR #196, opened 2015-10-09
update readme with HTTPError info

wwwdata:
this provides better information that we have a `HTTPError` struct that can be used as `error` return value. See #195


sharpner:
thx :+1:

PR #198, opened 2015-10-09
add missing return

### LINKED ISSUE #197: Unmarshaling fails with mixed case field names, e.g. ExpiryYear
I came across an issue I don't really know what to make of it. If I try to run [this test](https://gist.github.com/christianklotz/ac556f4037b13fc0dcf4) I get this:

```log
panic: invalid key "exp_year" in json. Cannot be assigned to target struct "ticket"
```

Interestingly the field exists, and the error only occurs if the `ticket` struct got the `Address` relationship and the `SetID` method is implemented in the `Address` struct. What I found is that in [`getFieldByTagName`, line 316](https://github.com/manyminds/api2go/blob/master/jsonapi/unmarshal.go#L316) it find the field successfully, however the returned result assigned in [`UnmarshalInto`, line 261](https://github.com/manyminds/api2go/blob/master/jsonapi/unmarshal.go#L261) is different than the result returned from the function. As soon as I don't use named return params it started working. I'm using `go 1.5.1`.

Can you reproduce this issue?

wwwdata:
this fixes #197

### DIFF
--- a/jsonapi/fixtures_test.go
+++ b/jsonapi/fixtures_test.go
@@ -355,6 +355,20 @@ func (s *SQLNullPost) SetID(ID string) error {
 	return nil
 }
 
+type RenamedPostWithEmbedding struct {
+	ID       string `jsonapi:"-"`
+	Another  string `jsonapi:"name=another"`
+	Field    string `jsonapi:"name=foo"`
+	Other    string `jsonapi:"name=bar-bar"`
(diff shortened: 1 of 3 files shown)


wwwdata:
i will add a test for this too and then push again

wwwdata:
ok, ready to review

PR #200, opened 2015-10-10
fix too early return when unmarshal embedded structs

### LINKED ISSUE #199: marshal-composition bug: cannot unmarshal a struct that contains both a dasherized tag and an embedded struct
This issue was introduced with the marshal-composition change.  In one of my libraries, I have a struct that contains both a field whose tag name is dasherized as well as an embedded struct. 

Prior to the marshal composition change, it worked fine.  After the change, it now raises an exception along the lines of 'invalid key "body-style" in json. Cannot be assigned to target struct "AutomobileResource"'.  If I remove the dash from the tag, then it's good to go.

I reproduced the issue in your package by embedding the Comment struct in the SimplePost struct like so:

https://github.com/manyminds/api2go/blob/master/jsonapi/fixtures_test.go#L68

```
type SimplePost struct {
	Comment
	ID          string `jsonapi:"-"`
	Title, Text string
	Internal    string `jsonapi:"-"`
	Size        int
	Created     time.Time `jsonapi:"name=create-date"`
	Updated     time.Time `jsonapi:"name=updated-date"`
}
```

After making that change and running the tests, four tests fail with "invalid key "create-date" in json. Cannot be assigned to target struct xxxxx" where xxxxx is SimplePost for 3 tests and TaggedPost for 1.

Its behavior is very similar to what was fixed in https://github.com/manyminds/api2go/issues/197.

wwwdata:
the search needs to go on if the field from the json was not found in the embedded struct and there are other fields. Now there is a check for that.

I changed the struct for the test, that triggered the error too. So we got that covered now.

This resolves #199

### DIFF
--- a/jsonapi/fixtures_test.go
+++ b/jsonapi/fixtures_test.go
@@ -356,12 +356,12 @@ func (s *SQLNullPost) SetID(ID string) error {
 }
 
 type RenamedPostWithEmbedding struct {
+	Embedded SQLNullPost
 	ID       string `jsonapi:"-"`
 	Another  string `jsonapi:"name=another"`
 	Field    string `jsonapi:"name=foo"`
 	Other    string `jsonapi:"name=bar-bar"`
 	Ignored  string `jsonapi:"-"`
-	Embedded SQLNullPost
 }
 
 func (p *RenamedPostWithEmbedding) SetID(ID string) error {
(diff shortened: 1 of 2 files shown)


sharpner:
It won't hurt to merge this, since its clearly a bug right? Even if we are not absolutely certain that it fixes #199

wwwdata:
yes

PR #201, opened 2015-10-10
Reference 3rd party example implementation

sharpner:


wwwdata:
big thanks to @hnakamur

PR #202, opened 2015-10-13
Major Refactoring: Fully typed structs that represent a jsonapi document

wwwdata:
### Background
It would be nice to finally get rid/ridder of the large untyped `map[string]interface{}` for our internal marshal/unmarshal methods. Especially regarding the #111 refactor unmarshaling task, a more statically typed internal data handling would be beneficial.

Now that all the payload is inside an attributes object in the json, this totally makes sense. When you look back how jsonapi started, this completely untyped approach seemed necessary, but now with the final specs, there is a better way!

### This PR will change a lot for api2go:
- We will be completely standard compatible. That means that we just use a normal `json.Marshal` and `json.Unmarshal` from the standard `encoding/json` package of go.
- Everything else is already inside interface methods and will be collected by executing these implemented methods.
- The main payload of a struct is just it's marshalled representation inside the `attributes` object in the jsonapi json. 

### Goals
- Finally get rid of bugs with specific special types
- Let the user marshal anything, because he can now just simply implement his own `Marshaler` and `Unmarshaler` interfaces and it will always work!

### Stuff that will break:
- Automatic renaming of fields (but maybe we can support that). It is advised to now use the `json:"yourName"` tag annotations
- Custom Content Marshalers that want to export anything else than JSON.. The new implementation is completely based upon the JSON interfaces from go.

### TODO
- [x] Jsonapi Package Marshal/Unmarshal
- [x] Api Package
- [x] Test package
- [x] Increase Test Coverage!
- [x] Documentation
- [x] Remove stuff from helpers.go 
- [x] Squash/Rebase to polish comits to be ready for review


jsonapi/data_structs.go
```diff
@@ -0,0 +1,35 @@
...
+	Type          string                  `json:"type"`
+	ID            string                  `json:"id"`
+	Attributes    map[string]interface{}  `json:"attributes"`
+	Relationships map[string]relationship `json:"relationships,omitempty"`
+	Links         links                   `json:"links,omitempty"`
+}
+
+type relationship struct {
+	Links      links              `json:"links,omitempty"`
+	DataToOne  relationshipData   `json:"data,omitempty"`
```
wwwdata: don't know if this works here, but jsonapi does an array for to-many and an object for to-one

sharpner:
ok so "update branch" merges the master into the PR. Sry for that :-) will rebase afterwards.

wwwdata:
yes, that feature is completely nuts from github :smile:

wwwdata:
already experimented with new structs and a Container struct that distinguishes from object and array payloads that are specified in the jsonapi spec. This is all possible with go Unmarshal interfaces. So in theory all this refactoring will work :smile:

wwwdata:
update: Refactoring of the code is mostly done. The commits are still a mess, but all the tests are working again! I mostly left them as they were, I just had to change them all to always test raw json `[]byte` as inputs, because that is what we are working now since we are 100% compatible to the api of `json.Marshaler` and `json.Unmarshaler` from the standard go package. :smile:

wwwdata:
I rebased my commits so they make more sense. Everybody that wants to do some code review or testing is welcome to do so!

Look at my changes in the example package, you should, if you are using our API route generation stuff, just need to change the annotations from `jsonapi` back to `json` and also give every exported field a name, because that's not done automatically anymore. 

To rename structs, it is needed to implement the `GetName() string` method from the `EntityNamer` interface.

Every super special whatever field type must be working now! Our json must be the same as generated with `json.Marshal` and `json.Unmarshal` from the go json package. We are just decorating that json result now with additional stuff from jsonapi. This should be a very  bullet proof approach.

sharpner:
I added a todo to remove helpers.go and its functions. :-)

wwwdata:
yep, delete all the unnecessary things!

wwwdata:
Current Status:
- We have to decide what to do with the `ContentMarshaler` interface. Because with the new architecture, only JSON is supported, nothing else. This breaks some stuff :D 
- And we now have structs that represent a jsonapi Document which are at the moment public and there even is a method `jsonapi.MarshalToStruct(data interface{}, ServerInformation) (Document, error)` to generate them instead of a `[]byte` of JSON.

sharpner:
Depends on #220

sharpner:
I already thought that content marshalers will be gone for now. If we ever decide to use them again we need a different solution anyway.

About the public structs: They are currently only public because they are public :) I cannot see any use case at the moment, and It makes it harder to get api stability imho. If they are private for now, we can see how it works out, and can even change them as long as the public api does not break.

If someone finds a valid use case for making them public, we can discuss it again :-)

But just for the sake of "someone could need it" its not worth it imho...

wwwdata:
ok @sharpner I will make the structs private and remove ContentMarshalers and then we are pretty much done and can merge.

wwwdata:
but we have to keep #217 in mind.

sharpner:
you are right. I would implement #217 as soon as this is merged. Fixing it in the current version makes no sense...

jsonapi/unmarshal.go
```diff
@@ -74,428 +72,153 @@ type EditToManyRelations interface {
...
-		if structType.Kind() == reflect.Ptr {
-			structType = structType.Elem()
-		}
-	}
-
-	if structType.Kind() != reflect.Struct {
-		return typeError
-	}
+var (
+	errInterface         = errors.New("target must implement UnmarshalIdentifier interface")
```
interlock: I often declare errors that are passed down to the consumer publicly so I can compare against the value. Works better than comparing the string `err.Error() = "..."` because it lets you change/clarify an error in the future without breaking code.

context.go
```diff
@@ -72,7 +73,7 @@ var _ APIContexter = &APIContext{}
 // ContextQueryParams fetches the QueryParams if Set
 func ContextQueryParams(c *APIContext) map[string][]string {
 	qp, ok := c.Get("QueryParams")
-	if ok == true {
+	if ok == false {
```
interlock: *face palm* how did I not catch that? Thanks for fixing it :-)

jsonapi/unmarshal.go
wwwdata: yes, good point, we should collect all the errors and make them public so a client can directly compare with them.

context.go
wwwdata: because you did not write a test :smile:

api.go
```diff
@@ -1154,3 +1101,58 @@ func handleError(err error, w http.ResponseWriter, r *http.Request, marshalers m
 
 	writeResult(w, []byte(marshaler.MarshalError(err)), http.StatusInternalServerError, contentType)
 }
+
+// TODO: this can also be replaced with a struct into that we directly json.Unmarshal
```
sharpner: what about this todo?

wwwdata:
I can refactor this too before merging.

context_test.go
```diff
@@ -41,4 +51,48 @@ var _ = Describe("Context", func() {
 			Expect(ok).To(BeFalse())
 		})
 	})
+
+	Context("Not yet implemented", func() {
```
sharpner: that context name looks weird :)

jsonapi/fixtures_test.go
```diff
@@ -66,13 +66,14 @@ func (u *User) SetID(stringID string) error {
...
-	Title, Text string
-	Internal    string `jsonapi:"-"`
-	Size        int
-	Created     time.Time `jsonapi:"name=create-date"`
-	Updated     time.Time `jsonapi:"name=updated-date"`
-	topSecret   string    `jsonapi:"name=top-secret"`
+	ID        string    `jsonapi:"-" json:"-"`
+	Title     string    `json:"title"`
+	Text      string    `json:"text"`
+	Internal  string    `jsonapi:"-" json:"-"`
```
sharpner: we should remove the jsonapi tags

jsonapi/unmarshal_test.go
```diff
@@ -16,229 +15,234 @@ import (
 var _ = Describe("Unmarshal", func() {
 	Context("When unmarshaling simple objects", func() {
 		t, _ := time.Parse(time.RFC3339, "2014-11-10T16:30:48.823Z")
-		singleJSON := []byte(`{"data":{"id": "1", "type": "simplePosts", "attributes": {"title":"First Post","text":"Lipsum", "Created": "2014-11-10T16:30:48.823Z"}}}`)
+		/*
```
sharpner: what's with the commented code?

context_test.go
wwwdata: it's on purpose because the tested functions do nothing and just fullfil the interface

jsonapi/fixtures_test.go
wwwdata: yes, I overlooked them

jsonapi/unmarshal_test.go
wwwdata: must be removed too :)

context_test.go
sharpner: ic thx

wwwdata:
ok, newest commit removes the `ContentMarshaler` completely. However we need to keep the new `Document` structs public because they are also used by `api.go` to include pagination links for example. If somebody else wants to add meta data, modify/add links or other stuff to a jsonapi document, he will need these structs too.

jsonapi/unmarshal.go
```diff
@@ -74,428 +72,153 @@ type EditToManyRelations interface {
...
 	}
 
-	// Read all the models
-	for _, m := range models {
-		data, ok := m.(map[string]interface{})
-		if !ok {
-			return errors.New("expected an array of objects under key data")
-		}
+	if ctx.Data.DataArray != nil {
+		targetType := reflect.TypeOf(target).Elem().Elem()
```
sharpner: this crashes with ```	curl -X POST http://localhost:31415/v0/chocolates -d '{"data" : [{"type" : "chocolates" , "attributes": {"name" : "Ritter Sport", "taste": "Very Good"}}]}'```

sharpner:
I am done reviewing this monster! :-D :+1 from me

PR #205, opened 2015-10-16
Cleaner error message by renamed unexported fields

sharpner:
The only way to check this that I could find was to use
recover.

If anyone has an idea how to check if a field is private beforehand I would
appreciate it deeply. :/

### DIFF
--- a/jsonapi/fixtures_test.go
+++ b/jsonapi/fixtures_test.go
@@ -72,6 +72,7 @@ type SimplePost struct {
 	Size        int
 	Created     time.Time `jsonapi:"name=create-date"`
 	Updated     time.Time `jsonapi:"name=updated-date"`
+	topSecret   string    `jsonapi:"name=top-secret"`
 }
 
 func (s SimplePost) GetID() string {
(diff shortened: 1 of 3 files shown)


wwwdata:
hm, normally the field should not have been found right? What exactly is the problem here? It finds the private field and then tries to set a value into it?

sharpner:
The problem is that due to the renaming flag it ignores the lookup and does not search for the field, but instead uses the name from the string and forces to set it.

wwwdata:
ah, yes make sense so it finds the private field with the tag. So can we just check if the field is private anyway and then not set it even if it has a tag?

sharpner:
How do you check if a field is unexported? It's impossible I read the whole fucking docu :D

wwwdata:
we did something similar in the marshaler and checked it with
```
!field.CanInterface()
```

we cannot do that here?

sharpner:
I swear, that's the first thing I tried :D Must have had some error somewhere else in the code previously.

wwwdata:
ok ok, that's why we have these damn pull requests for, looki looki

sharpner:
:+1: =)

wwwdata:
ok if tests go through then :shipit:

PR #207, opened 2015-10-21
fall back to json tag for name on fields

### LINKED ISSUE #206: Fallback to `json` field tags for Marshal/Unmarshal?
Our models are a little more universal than just jsonapi and already have tags for the json encoder:

```go
type User struct {
 Id int64 `json:"id"`
 Name string `json:"name"`
}
```

`Jsonify` https://github.com/manyminds/api2go/blob/master/jsonapi/helpers.go#L62 doesn't match our specs.

Adding duplicate data `jsonapi:name="id";json:"id"` can be a source of simple, but frustrating mistakes. e.g. typo in one but not the other.

I propose we fallback to using a `json` tag if it exists to get the name in func `GetTagValueByName` https://github.com/manyminds/api2go/blob/master/jsonapi/helpers.go#L86

This could open the door up to get omitempty and other tags that appear to be mirrored in jsonapi from json.

interlock:
Implements falling back to `json` field tags if there is no `jsonapi` tag.

closes #206

### DIFF
--- a/jsonapi/helpers.go
+++ b/jsonapi/helpers.go
@@ -86,7 +86,16 @@ func Singularize(word string) string {
 func GetTagValueByName(tfield reflect.StructField, name string) string {
 	str := tfield.Tag.Get("jsonapi")
 	if str == "" {
-		return ""
+		str = tfield.Tag.Get("json")
+		if str == "" || str == "-" {
+			return ""
+		}
+
+		if idx := strings.Index(str, ","); idx != -1 {
+			return str[:idx]
+		}
+		return str
+
 	}
 
 	tags := strings.Split(str, ";")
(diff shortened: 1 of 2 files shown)


jsonapi/helpers.go
```diff
@@ -86,7 +86,16 @@ func Singularize(word string) string {
 func GetTagValueByName(tfield reflect.StructField, name string) string {
 	str := tfield.Tag.Get("jsonapi")
 	if str == "" {
-		return ""
+		str = tfield.Tag.Get("json")
+		if str == "" || str == "-" {
+			return ""
+		}
+
+		if idx := strings.Index(str, ","); idx != -1 {
```
sharpner: this would only read the first setting. If the tag would be `json:"foo=bar,name=blub"` it would return bar I think.

interlock: That is actually the spec for the json encode tags. They do not follow the same convention as api2go.

sharpner: Didn't know that :D

sharpner:
discussion in #206

sharpner:
can I close this @interlock ?

PR #215, opened 2015-10-29
Show UnmarshalInto dropping attributes set to null.

dennisfaust:
Below is the output of running this integration test and accompanying debug output. This PR demonstrates how in api.go's handleUpdate function the context (ctx) is unmarshaled using encoding/json and correctly contains the attribute named value being set to null. See the Context: below where value:interface{}(nil) is correct. But then after going through jsonapi.UnmarshalInto the nil value has been dropped and the test fails because the attribute was not updated.
```
Context: map[string]interface {}{"data":map[string]interface {}{"type":"posts", "id":"1", "attributes":map[string]interface {}{"title":"New Title", "value":interface {}(nil)}}}

Updated Obj: &{1 New Title {{2 true}} 0xc8205897e0 [{1 This is a stupid post!}] []}

------------------------------
• Failure [0.003 seconds]
RestHandler
/Users/dennis/Sites/go/src/github.com/api2go/api_test.go:1599
  when handling requests for pointer resources
  /Users/dennis/Sites/go/src/github.com/api2go/api_test.go:982
    Updating
    /Users/dennis/Sites/go/src/github.com/api2go/api_test.go:975
      UPDATEs correctly using null.* values [It]
      /Users/dennis/Sites/go/src/github.com/api2go/api_test.go:853

      Expected
          <null.Float>: {
              NullFloat64: {Float64: 2, Valid: true},
          }
      to equal
          <null.Float>: {
              NullFloat64: {Float64: 0, Valid: false},
          }

      /Users/dennis/Sites/go/src/github.com/api2go/api_test.go:852
```

### DIFF
--- a/jsonapi/unmarshal.go
+++ b/jsonapi/unmarshal.go
@@ -267,10 +267,11 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type, 
 
 					value := reflect.ValueOf(attributeValue)
 
+					if !field.CanInterface() {
+						return fmt.Errorf("field not exported. Expected field with name %s to exist", fieldName)
+					}
+
 					if value.IsValid() {
-						if !field.CanInterface() {
-							return fmt.Errorf("field not exported. Expected field with name %s to exist", fieldName)
-						}
 
 						plainValue := reflect.ValueOf(attributeValue)
 
@@ -298,6 +299,15 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type, 
 									return fmt.Errorf("Could not set field '%s'. %s", fieldName, err.Error())
 								}
 							}
+						}
+					} else {
+						// Set the field to its zero value if its type is from the zero or null package.
+						if extType := field.Type(); extType != nil && extType.Kind().String() == "struct" {
+							fieldSource := extType.Field(0).Name
+							if strings.HasPrefix(fieldSource, "Null") || strings.HasPrefix(fieldSource, "Zero") || strings.HasPrefix(fieldSource, "Time") {
+								field.Set(reflect.Zero(field.Type()))
+							}
+
 						}
 					}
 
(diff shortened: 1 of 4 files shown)


wwwdata:
especially with these null types there is a lot of manual stuff happening at the moment because the unmarshalling is currently done into a `map[string]interface{}` and not directly into the target struct. I will start this weekend with some refactoring #202. Like I described there. My approach, now that all the payload data is inside the `attributes` object in the json, is to directly marshal/unmarshal from a struct into that. So no more manual map magic that leads to unexpected behavior. The goal will be to be 100% json marshal/unmarshal compatible and just decorate that with everything that's needed from jsonapi spec.

dennisfaust:
I added handling for null and zero values into UnmarshalInto. It works for my tests but the whole thing is less than ideal.

dennisfaust:
@wwwdata in the meantime would you consider merging this PR?

jsonapi/unmarshal.go
```diff
@@ -299,6 +300,16 @@ func UnmarshalInto(input map[string]interface{}, targetStructType reflect.Type,
 								}
 							}
 						}
+					} else {
+						//fmt.Printf("\nfieldname: %v, fieldType: %#v, type: %#v, value: %#v, IsValid: %v\n", fieldName, fieldType, field.Type().Field(0).Name, value, value.IsValid())
```
sharpner: these lines should be removed before merging

dennisfaust:
Comment removed per @sharpner.

wwwdata:
Please Squash your commits and rebase them into our master. We can merge it then. I guess I will also finish my huge refactoring at the end of this or next week. After that such issues won't come up anymore.

dennisfaust:
Rebased to master and squashed.

wwwdata:
Thanks a lot for your contribution.

PR #219, opened 2015-11-25
always take the type our of the ReferenceID struct

### LINKED ISSUE #218: Allow different types in one many-relationship
In my case I have an ordered list of resources of different types.
I want to return them in a single reference array.

In my GetReferencedIDs I dynamically set the different types:
```
// GetReferencedIDs MarshalLinkedRelations
func (p Path) GetReferencedIDs() []jsonapi.ReferenceID {
	result := []jsonapi.ReferenceID{}

	for _, elem := range p.PathElements {
		var returnType string

		switch elem.Type {
		case "Question":
			returnType = "questions"
		case "Idea":
			returnType = "ideas"
		case "Problem":
			returnType = "problems"
		}

		result = append(result, jsonapi.ReferenceID{
			ID:   elem.ID,
			Type: returnType,
			Name: "path-elements",
		})
	}

	return result
}
```

But the api returns in my ```path-elements``` field for every resource the same type and ignores the various types.

Example output:
```
"relationships": {
            "path-elements": {
                "data": [
                    {
                        "id": "f48aead1-5721-4bf0-8801-301006013b51",
                        "type": "questions"
                    },
                    {
                        "id": "20d3683e-9982-410d-a77a-7728d4f625f9",
                        "type": "questions"
                    },
                    {
                        "id": "e1d923cd-0882-4978-8f1d-1f5409094e71",
                        "type": "questions"
                    },
                    {
                        "id": "713536d3-7ce6-4035-99ec-c73310e60dc6",
                        "type": "questions"
                    },
                    {
                        "id": "068c57b6-a0d2-4354-b6e6-71af7f12bfce",
                        "type": "questions"
                    }
                ],
                "links": {
                    "related": "/v1/paths/f48aead1-5721-4bf0-8801-301006013b51/path-elements",
                    "self": "/v1/paths/f48aead1-5721-4bf0-8801-301006013b51/relationships/path-elements"
                }
            }
        },
```

wwwdata:
when marshalling a toMany relationship with the same name but with different types in it, it was previously not possible to set the type for every referenced ID because it was assumed that it would always be of the same type. This is now possible.

Fixes #218

### DIFF
--- a/jsonapi/marshal.go
+++ b/jsonapi/marshal.go
@@ -245,7 +245,6 @@ func getStructRelationships(relationer MarshalLinkedRelations, information Serve
 	}
 
 	for name, referenceIDs := range sortedResults {
-		referenceType := referenceIDs[0].Type
 		relationships[name] = map[string]interface{}{}
 		// if referenceType is plural, we need to use an array for data, otherwise it's just an object
 		if Pluralize(name) == name {
@@ -254,7 +253,7 @@ func getStructRelationships(relationer MarshalLinkedRelations, information Serve
 
 			for _, referenceID := range referenceIDs {
 				data = append(data, map[string]interface{}{
-					"type": referenceType,
+					"type": referenceID.Type,
 					"id":   referenceID.ID,
 				})
 			}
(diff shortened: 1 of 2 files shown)


sharpner:
LGTM

mantenie:
:+1: The test case matches exactly that what I wanted.

PR #223, opened 2015-12-11
Update readme to direct new users to v1

sharpner:
we should hint v1 instead of v0 so that new users use the new codebase


wwwdata:
gopkg does not re-write sub packages to the correct include path. we will now have to do the following:

* recommend master branch for up-to-date users
* fix imports manually in our v0 branch to gopkg v0

sharpner:
v0 is fixed

sharpner:
fixed default readme, @wwwdata could you take a look?

PR #224, opened 2015-12-14
nil Marshals into json object with null

### LINKED ISSUE #222: how to respond with json {"data": null} when GET single non-existing resource?
Firstly, thanks for the great work - the code is a pleasure to read and a great source for learning to me!

jsonapi specification says:

    A server MUST respond to a successful request to fetch an individual resource with a resource object or null provided as the response document's primary data.
    null is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn't currently.

Example:

    HTTP/1.1 200 OK
    Content-Type: application/vnd.api+json
    {
        "links": {
            "self": "http://example.com/bills/108"
        },
        "data": null
    }

How do i construct such a response in my FindOne() CRUD handler?
I have tried to return both:

    return &Response{Res: nil, Code: http.StatusOK}, nil

and

    return &Response{Code: http.StatusOK}, nil

from the handler and get a "nil cannot be marshalled" error.

My Response type is a trivial:

    type Response struct {
  	  Res  interface{}
	  Code int
    }

    func (r Response) Result() interface{} {
        return r.Res
    }

What am i missing or doing wrong?

wwwdata:
this resolves #222

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -288,6 +288,11 @@ func (s *fixtureSource) PaginatedFindAll(req Request) (uint, Responder, error) {
 }
 
 func (s *fixtureSource) FindOne(id string, req Request) (Responder, error) {
+	// special test case for nil document
+	if id == "69" {
+		return &Response{Res: nil}, nil
+	}
+
 	if p, ok := s.posts[id]; ok {
 		if s.pointers {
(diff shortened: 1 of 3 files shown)


sharpner:
ty!

PR #225, opened 2015-12-14
patch to fix update not using the actual server info 

### LINKED ISSUE #216: `handleUpdate` should not pass an empty information{} to Marshaller
If the resource in question implements the relationship related interfaces this causes a it to attempt to run functions on the empty `information{}` resulting in a `panic()`.

Fix is to pass along the info as the other handlers do and pass that to the marshaller.

https://github.com/manyminds/api2go/blob/master/api.go#L728

wwwdata:
this is a cherry pick from picatic and resolves #216


sharpner:
nice, should we add a testcase?

wwwdata:
yes, a test is needed here too

sharpner:
nice thx!

PR #227, opened 2015-12-15
Remove content marshalers from the api2go with gin example

hnakamur:
This change is the same as in https://github.com/manyminds/api2go-adapter/pull/4/files


wwwdata:
thx!

PR #228, opened 2015-12-16
Add issuestats badges

sharpner:


wwwdata:
Gief all them badges!

Issue #234, opened 2016-01-19
Updating Resource relationship throws 406 attributes is mandatory


tanmayambre:
Hello 
I refer to you to the following specification. 
http://jsonapi.org/format/#crud-updating-resource-relationships.

I am trying to update relationship via curl using the spec mentioned. Note that there are no "attributes"  to be updated in the third curl request only a "relationships".

I used the provided crud_example.go to test this. 
```
#create a user
curl -X POST http://localhost:31415/v0/users -d '{"data" : {"type" : "users" , "attributes": {"user-name" : "marvin"}}}'

#create a chocolate
curl -X POST http://localhost:31415/v0/chocolates -d '{"data" : {"type" : "chocolates" , "attributes": {"name" : "Ritter Sport", "taste": "Very Good"}}}'

#update sweet relationship of an user
curl -X PATCH http://localhost:31415/v0/users/1 -d '{"data":{"type":"users","id":"1","relationships":{"sweets":{"data":{"type":"chocolates","id":"1"}}}}}'

```
The last request should have returned a 204 No Content. Instead I get the following 406 error.
```
{"errors":[{"status":"406","title":"missing mandatory attributes object"}]}
```

Please advice if I should be expecting the message or is a bug. It seems that this has changed recently.

wwwdata:
you are right, the update method now always expects an `attributes` object. We forgot the use case where we only want to replace all the `relationships` of a resource. 

if `attributes` is added, it works 
```
curl -X PATCH http://localhost:31415/v0/users/1 -d '{"data":{"type":"users","id":"1","attributes": {},"relationships":{"sweets":{"data":[{"type":"chocolates","id":"1"}]}}}}'
```

apart from that, you can also use the dedicated routes to manipulate relationships which should be working correctly (because we have tests for them). 

So, for example this is working as well:
```
curl -X DELETE http://localhost:31415/v0/users/1/relationships/sweets -d '{"data":[{"type":"chocolates","id":"1"}]}'
```

Or add one again:
```
curl -X POST http://localhost:31415/v0/users/1/relationships/sweets -d '{"data":[{"type":"chocolates","id":"1"}]}'
```

but good catch! This is a bug because the specification says it should be working.

wwwdata:
we implemented it wrong, it's totally fine to leave out an `attributes` object:

> A resource object MUST contain at least the following top-level members:
> - id
> - type
> Exception: The id member is not required when the resource object originates at the client and represents a new resource to be created on the server.

[...]

PR #235, opened 2016-01-22
Fix typo in error message

gabesullice:


wwwdata:
ah yes it's a typo but this check is wrong anyway see #234 
it's totally fine to create a jsonapi object and only fill out the `type` field and nothing else. 

But I will merge this just to delete it again, because it's funny :smile:

PR #241, opened 2016-02-12
Fix double forward slash if no prefix was set

### LINKED ISSUE #240: Location header returns 2 forward slash "//users/1" Is this correct behaviour?
In crud_example.go code I stared the NewApi("") with no prefix as below 
```	
    port := 31415
	// api := api2go.NewAPIWithResolver("v0", &resolver.RequestURL{Port: port})
	api := api2go.NewAPI("")
	userStorage := storage.NewUserStorage()
```

Made a POST request as follows
```
curl -X POST http://localhost:31415/users -d '{"data" : {"type" : "users" , "attributes": {"user-name" : "marvin"}}}' -v
...
< Location: //users/1
...
```
I was expecting the Location header to be `/users/1`. 
Currently it adds a extra forward slash. `//users/1`
Is this the correct behaviour?

wwwdata:
this fixes #240

### DIFF
--- a/api.go
+++ b/api.go
@@ -585,7 +585,11 @@ func (res *resource) handleCreate(c APIContexter, w http.ResponseWriter, r *http
 		return fmt.Errorf("Expected one newly created object by resource %s", res.name)
 	}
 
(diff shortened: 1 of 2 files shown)


sharpner:
awesome, thank you

Issue #243, opened 2016-02-24
handleRead hardcodes http.StatusOk as status code


tochiai:
It is not possible to set a response status code via the [`response`](https://github.com/manyminds/api2go/blob/master/api.go#L465) returned by `FindOne`; the status code is set to [`http.StatusOk`](https://github.com/manyminds/api2go/blob/master/api.go#L471). Would it be possible to add logic in `handleRead` to try to use `response.StatusCode()`?

sharpner:
The specification of jsonapi only specifies 200 for OK and 404 for not found. http://jsonapi.org/format/#fetching-resources

Could you describe your problem a little bit?

tochiai:
I was trying to follow this pattern for asynchronous processing 
http://jsonapi.org/recommendations/#asynchronous-processing

Glossing over some name changes, I'm basically trying to implement the part where a 303 is returned for `GET /photos/queue-jobs/5234`.

My reading of the spec allows for other status codes when fetching a resource, what do you think?
http://jsonapi.org/format/#fetching-resources-responses-other

wwwdata:
you are right, I looked at the code and this is a bug. 

Theoretically you should be able to set a status code in the `Responder`, but in our `FindOne` method of the api, we always set a 200 which is not right. We have to change this.

sharpner:
Regardless of wether it is a bug or a feature, the next problem will be to implement the "see other" location header. But I'm working on it. If you can live without the self link which is handled in #99 it should be done soon. :-)

sharpner:
@tochiai I implemented most of the basics in #244 if you want you can test it. But its not yet possible to perfectly implement it.

PR #244, opened 2016-02-25
Asynchronous processing

sharpner:
TODOs:
- [x] Nice handling of url creation
- [ ] Find a clean solution to handle subtrees e.g. /jobs/queues/1 like in the spec

### DIFF
--- a/api.go
+++ b/api.go
@@ -468,6 +468,15 @@ func (res *resource) handleRead(c APIContexter, w http.ResponseWriter, r *http.R
 		return err
 	}
 
+	if response.StatusCode() == http.StatusAccepted {
+		return res.respondWith(response, info, http.StatusAccepted, w, r)
+	}
+
+	if seeOther, ok := response.(SeeOtherResponder); ok {
+		w.Header().Add("Location", info.GetBaseURL())
+		return res.respondWith(seeOther, info, http.StatusSeeOther, w, r)
+	}
+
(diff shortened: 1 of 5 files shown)


sharpner:
So I kinda got this to work, but the tricky part is now the correct handling.
The actual flow of asynch operations would be:
### POST /photos
Create gets called, but it does not create a photo, instead it creates a temporary handle. In the example
it creates /photos/queue-jobs/5234 that only contains a "status" attribute. 
Optionally you get a *Content-Location* Header with the optional full resource identifier.

### GET /photos/queue-jobs/5234
This call will return 303 See Other when its done. Additionally it contains a *Location* Header back to the actual Photo that got created. E.g. *Location: https://example.com/photos/4577*
No documentation as to what is if the process is not yet finished.

IMHO api2go does not yet have the functionality to implement such an behaviour. A user can probably fake it after this PR gets merged (without the optional location header) but to implement the suggestion 1:1 we would need a way to respond to Create differently than either Succeed or Error.

What do you think @wwwdata?

sharpner:
RFC for #243

wwwdata:
so far this looks good in my opinion

sharpner:
so we are going to merge this for now?

wwwdata:
It cannot hurt to merge it but we should mention something about it in the Readme as well

PR #247, opened 2016-05-15
Added InitializeObject hook

256dpi:
Hi there!

First of all thanks for this amazing library!

I'm working on a small very opinionated framework that uses struct tags to simplify the declaration of models. I first thought this might get very complicated with `api2go` as it uses `reflect` to create new objects internally. But in fact I learned that most actions go either through `crud.Find()` or `crud.FindAll()` which allows to setup the object before `jsonapi.Marshal()` or `jsonapi.Unmarshal()` is called. The situation that does not allow such access is on a PATCH request. An object gets created with `reflect` right before passing it to `jsonapi.Unmarshal()`.

TL;DR:
This pull request adds an optional method that can be implemented to get access to the dynamically created object in a PATCH request before `jsonapi.Unmarshal()` is called. This is necessary to give context to the models in higher frameworks that abstract the `jsonapi` interfaces.

Please feel free to change the name of the interface and method. ;)

### DIFF
--- a/api.go
+++ b/api.go
@@ -562,6 +562,12 @@ func (res *resource) handleCreate(c APIContexter, w http.ResponseWriter, r *http
 	}
 	newObj := reflect.New(resourceType).Interface()
 
+	// Call InitializeObject if available to allow implementers change the object
+	// before calling Unmarshal.
+	if initSource, ok := res.source.(ObjectInitializer); ok {
+		initSource.InitializeObject(newObj)
+	}
+
 	err = jsonapi.Unmarshal(ctx, newObj)
 	if err != nil {
 		return NewHTTPError(nil, err.Error(), http.StatusNotAcceptable)
--- a/api_interfaces.go
+++ b/api_interfaces.go
@@ -48,6 +48,14 @@ type FindAll interface {
 	FindAll(req Request) (Responder, error)
 }
 
+// The ObjectInitializer interface can be implemented to have the ability to change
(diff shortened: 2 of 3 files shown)


wwwdata:
Thanks for the PR. Yes what you are saying makes sense. Im just confused that you mentioned `PATCH`. You mean `POST` because you called the potential interface implementation in the Create api route, right? Which makes totally sense because in that, we create a new instance of a struct and then unmarshal into it, which would be the undesired behaviour you are describing.

I currently don't really understand why go 1.4 fails in travis, but I guess it's some golint issues.

I like the interface name, but please squash the `fix typo` commit into your first, I'll check that the test are working and then we can merge it :)

wwwdata:
https://github.com/golang/lint/issues/198
> Per the README, golint nowadays requires Go 1.5 or later.

ok, hm, let's ditch that old version 1.4 then :D

sharpner:
I would love if you could add a test using the newly introduced behaviour, and additionally shouldn't the interface be something like ```ObjectInitializer``` to keep golangs interface guidelines?

256dpi:
Thanks for the fast responses.

@wwwdata Ups, yes I mean `POST` of course.
@sharpner You're right, that makes more sense. I will add a test in a moment.

wwwdata:
once #248 is merged you can just rebase on it for the tests to work

wwwdata:
:+1: just one more rebase before merge

256dpi:
Thanks!

Issue #250, opened 2016-06-14
Ability to specify whether relationships are plural


jlburkhead:
Currently relationships are determined to be to-one or to-many by [checking if the `ReferenceID.Name` field is plural](https://github.com/manyminds/api2go/blob/master/jsonapi/marshal.go#L216). `Pluralize` uses https://github.com/gedex/inflector to determine the plural name which is a port of the CakePHP inflector library. There are some edge cases which are pluralized incorrectly in the [CakePHP inflector](http://inflector.cakephp.org/criteria) and thus also in `github.com/gedex/inflector` (I noticed this when I tried to make a to-many relationship with the name `criteria`). 

It seems like the best solution is to make it possible to override the default behavior and specify a relationship as either to-one or to-many. I did a quick scan of the jsonapi spec and couldn't find anything that requires a relationships' pluralization to indicate whether it is to-one or to-many, so allowing this to be overridden wouldn't create the problem where you could unknowingly break the spec.

Can turn this into a PR, add tests and clean up if you want to implement this.
```diff
diff --git a/jsonapi/marshal.go b/jsonapi/marshal.go
index 9614ddc..cde789c 100644
--- a/jsonapi/marshal.go
+++ b/jsonapi/marshal.go
@@ -8,6 +8,14 @@ import (
        "strings"
 )
 
+type RelationshipType int
+
+const (
+       DEFAULT RelationshipType = iota
+       TOONE
+       TOMANY
+)
+
 // MarshalIdentifier interface is necessary to give an element
 // a unique ID. This interface must be implemented for
 // marshal and unmarshal in order to let them store
@@ -19,9 +27,10 @@ type MarshalIdentifier interface {
 // ReferenceID contains all necessary information in order
 // to reference another struct in jsonapi
 type ReferenceID struct {
-       ID   string
-       Type string
-       Name string
+       ID           string
+       Type         string
+       Name         string
+       Relationship RelationshipType
 }
 
 // Reference information about possible references of a struct
@@ -191,6 +200,13 @@ func marshalData(element MarshalIdentifier, information ServerInformation) (*Dat
        return result, err
 }
 
+func isToMany(referenceID ReferenceID) bool {
+       if referenceID.Relationship == DEFAULT {
+               return Pluralize(referenceID.Name) == referenceID.Name
+       }
+       return referenceID.Relationship == TOMANY
+}
+
 // getStructRelationships returns the relationships struct with ids
[...]

PR #251, opened 2016-06-15
add ability to override default plural check

jlburkhead:
This is for https://github.com/manyminds/api2go/issues/250

### DIFF
--- a/jsonapi/marshal.go
+++ b/jsonapi/marshal.go
@@ -8,6 +8,18 @@ import (
 	"strings"
 )
 
+// RelationshipType is used to specify whether the relationship with referenced objects are to-one or to-many
+type RelationshipType int
+
+const (
+	// DefaultRelationship guesses the relationship type based on the pluralization of the reference name
+	DefaultRelationship RelationshipType = iota
+	// ToOneRelationship forces the relationship to be to-one
+	ToOneRelationship
+	// ToManyRelationship forces the relationship to be to-many
+	ToManyRelationship
+)
+
 // MarshalIdentifier interface is necessary to give an element
 // a unique ID. This interface must be implemented for
 // marshal and unmarshal in order to let them store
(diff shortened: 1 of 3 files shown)


wwwdata:
lgtm. It will also not crash if no relationship type was specified and default to the `DefaultRelationship`, right? What do you think @sharpner?

jlburkhead:
correct. the zero value of `RelationshipType` is `DefaultRelationship`. I can add a test for that, but I figured it was already implicitly being tested by none of the existing tests failing

wwwdata:
yes, let's wait for our second pr review and wen can merge :) thanks a lot

jsonapi/marshal.go
```diff
@@ -213,7 +234,7 @@ func getStructRelationships(relationer MarshalLinkedRelations, information Serve
 		relationships[name] = Relationship{}
 		// if referenceType is plural, we need to use an array for data, otherwise it's just an object
 		container := RelationshipDataContainer{}
-		if Pluralize(name) == name {
+		if isToMany(referenceIDs[0].Relationship, referenceIDs[0].Name) {
```
sharpner: we should check the length of referencedIDs before accessing `0`, could possibly be out of range.

sharpner:
awesome :-) thanks for the PR! this looks really good.

jsonapi/marshal.go
jlburkhead: `referencedIDs` has to have at least 1 element since `sortedResults` is populated [here](https://github.com/jlburkhead/api2go/blob/75f0e601dcdbfd113d0c6a8c29b6cba3f0b7f480/jsonapi/marshal.go#L222). `referenceIDs[0]` is also accessed [here](https://github.com/jlburkhead/api2go/blob/75f0e601dcdbfd113d0c6a8c29b6cba3f0b7f480/jsonapi/marshal.go#L248) without checking length

Issue #260, opened 2016-08-27
PATCHing a reference that is not loaded should return a 403


256dpi:
Hi, I found a small inconsistency while working on [fire](https://github.com/256dpi/fire).

Let's say I have a model that exposes a HasMany relationship that is not loaded:

```go
// Post
jsonapi.Reference{
  Type: "comments",
  Name: "comments",
  IsNotLoaded: true,
}
```

This will omit the `data: []` attribute from responses to the following URL to fetch the relationship:

```
GET /posts/1/relationships/comments
```

First of all: Is this even allowed by the spec or should this call return an error?

In any case the said relationship can currently also be updated via a PATCH request that includes an array of resource links. However, since the references aren't even loaded the API probably does not allow to set them either. In that case we must return an error. The spec suggest to return a `403 Forbidden` if the method is not allowed: http://jsonapi.org/format/#crud-updating-relationships.

I'm even wondering if the PATCH route should be generated at all?

wwwdata:
Hm, indeed this is a problem. The main problem is that we just call the `FindOne` method of the resource which in this case most likely behaves like a normal request with embedded relations.
Currently this could be avoided by checking the url path in the request which should include the url part `relationships` but a more dedicated resource method to fetch relationships would be better, right?

We did it in this way at first to just simplify the methods needed to be implemented for CRUD and make this relationship route to be automatically available. But also in terms of database optimizations, this approach is not very good. I would suggest to optionally add another method via interface, check for this and then call this dedicated resource method to only fetch the specific relationship data which then also can be done in a different database query. Would you like to work on this?

256dpi:
Yes, a method on the resource that is responsible for fetching such relations would be the way to go.

I'll look into submitting a PR for fixing the bug first.

PR #261, opened 2016-08-28
Some improvements

256dpi:
To get started with #260 I made myself a little more familiar with the internals of this package and came up with some small improvements.

I would also like to rewrite some of the functions documentations and optimize the code a little more. Would that be ok for you?

### DIFF
--- a/jsonapi/marshal.go
+++ b/jsonapi/marshal.go
@@ -8,28 +8,28 @@ import (
 	"strings"
 )
 
-// RelationshipType is used to specify whether the relationship with referenced objects are to-one or to-many
+// RelationshipType specifies the type of a relationship.
 type RelationshipType int
 
+// The available relationship types.
+//
+// Note: DefaultRelationship guesses the relationship type based on the
+// pluralization of the reference name.
 const (
-	// DefaultRelationship guesses the relationship type based on the pluralization of the reference name
 	DefaultRelationship RelationshipType = iota
-	// ToOneRelationship forces the relationship to be to-one
 	ToOneRelationship
-	// ToManyRelationship forces the relationship to be to-many
 	ToManyRelationship
 )
 
-// MarshalIdentifier interface is necessary to give an element
-// a unique ID. This interface must be implemented for
-// marshal and unmarshal in order to let them store
-// elements
+// The MarshalIdentifier interface is necessary to give an element a unique ID.
+//
+// Note: The implementation of this interface is mandatory.
 type MarshalIdentifier interface {
 	GetID() string
 }
 
-// ReferenceID contains all necessary information in order
-// to reference another struct in jsonapi
+// ReferenceID contains all necessary information in order to reference another
+// struct in JSON API.
 type ReferenceID struct {
 	ID           string
 	Type         string
@@ -37,10 +37,12 @@ type ReferenceID struct {
 	Relationship RelationshipType
 }
 
-// Reference information about possible references of a struct
-// If IsNotLoaded is set to true, the `data` field will be omitted and only the `links` object will be
-// generated. You should do this if there are some references, but you do not want to load them.
-// Otherwise, if IsNotLoaded is false and GetReferencedIDs() returns no IDs for this reference name, an
+// A Reference information about possible references of a struct.
+//
+// Note: If IsNotLoaded is set to true, the `data` field will be omitted and only
+// the `links` object will be generated. You should do this if there are some
+// references, but you do not want to load them. Otherwise, if IsNotLoaded is
+// false and GetReferencedIDs() returns no IDs for this reference name, an
 // empty `data` field will be added which means that there are no references.
 type Reference struct {
 	Type         string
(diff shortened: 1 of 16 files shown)


wwwdata:
Hi. Of course it's ok if you improve and optimize stuff! Thank you very much. Please tell us when you are finished and This can be reviewed and merged. I have seen you are pushing more and more commits. And also squash your commits where it makes sense when you are done please. Commits like `cleanup` and so on for example.

jsonapi/marshal.go
```diff
@@ -285,67 +287,68 @@ func getStructRelationships(relationer MarshalLinkedRelations, information Serve
...
 
 // helper method to generate URL fields for `links`
 func getLinksForServerInformation(relationer MarshalLinkedRelations, name string, information ServerInformation) *Links {
-	links := &Links{}
-
-	if information != serverInformationNil {
-		prefix := strings.Trim(information.GetBaseURL(), "/")
-		namespace := strings.Trim(information.GetPrefix(), "/")
-		structType := getStructType(relationer)
+	if information == nil {
```
sharpner: this won't work because the types don't match anymore IIRC, we should keep the `serverInformationNil` for this check.

256dpi: As ServerInformation is an interface, the parameter information should either point to an implementor or be nil. I don't see the need for `serverInformationNil`. Maybe I'm missing something here.

256dpi:
I'm finished with rephrasing the documentation and formatting the code. The last commit turned out to be a little bigger than I expected (I got a little carried away with formatting all the JSON in the test files...) I hope you don't mind.

I'll not add anymore commits to this PR and start working towards #260 on a new branch.

In that sense this PR is ready to be reviewed.

sharpner:
Thanks for your hard work! I'll Review this asap

jsonapi/marshal.go
```diff
@@ -113,107 +118,108 @@ func MarshalToStruct(data interface{}, information ServerInformation) (Document,
...
 
-	return &includedElements
+	return includedElements, nil
 }
 
-func marshalData(element MarshalIdentifier, information ServerInformation) (*Data, error) {
-	var err error
-	result := &Data{}
-
+func marshalData(element MarshalIdentifier, data *Data, information ServerInformation) error {
```
wwwdata: this is better because we can already write into a pre allocated `Data` struct instead of creating and returning a new one, right?

jsonapi/marshal.go
wwwdata: I'm not sure anymore why we did it in that weird way, and  a simple nil check is always better but if all tests are running it should be working.

jsonapi/unmarshal.go
```diff
@@ -7,80 +7,77 @@ import (
...
+// Deletes comments that belong to post with ID 1.
+// The DeleteToManyIDs method will be called.
 type EditToManyRelations interface {
 	AddToManyIDs(name string, IDs []string) error
 	DeleteToManyIDs(name string, IDs []string) error
 }
 
-var (
-	// ErrInterface occurs if the target struct did not implement the UnmarshalIdentifier interface
-	ErrInterface = errors.New("target must implement UnmarshalIdentifier interface")
```
wwwdata: Hm, we made this a variable and exported from the package because we wanted that somebody from the outside is able to check against this error variable. But I guess for this error it does not really matter. However for the `ErrNoType` it could be useful right? Because that actually checks that the `type` field is missing int he JSON that needs to be unmarshaled.

wwwdata:
@256dpi thank you very much for all this work! I really like that you cleaned up so much and wrote way better comments! When @sharpner has reviewed too, we can merge. The only thing we should be careful with is the removal of the eported error variables, if somebody depended on them, their update of the library would break because they are no longer available.

jsonapi/marshal.go
256dpi: Exactly.

jsonapi/unmarshal.go
256dpi: I removed the errors to be in sync with all the other methods. Marshal for example does not return an exported error. I agree that exporting them might be useful, but then it should happen in a concise way.

256dpi:
@wwwdata Thanks!

About the errors: Maybe we can add them back in a concise way in another PR?

wwwdata:
@256dpi I agree, we should go through all errors and decide which of them would make sense to be exported.

jsonapi/marshal.go
wwwdata: I think this shows the complication with nil and interface parameters: https://play.golang.org/p/W8VNDB4_9m

wwwdata: but it should be fine now because it's only called with a real `nil` or the "Implementation" of the interface now, right?

256dpi: Found some more info here: https://golang.org/doc/faq#nil_error.

I think so. I mean if the caller is funny and passes one of those weird nil typed interfaces, he should also be responsible for dealing with the panic. I'm not aware that it is a known go pattern to check all interface parameters to be "really" nil.

256dpi: I can only find one situation:
https://github.com/manyminds/api2go/blob/5207f05d44090999ebf2e09b3e66566484c4dbee/api.go#L227

It might be possible that `info = &api.info` can return one such a nil typed interface if it is not set?

256dpi: I think it should be fine as `api.info` should anyways be always set to something at it is a concrete type.

wwwdata: all the `New` methods that create an `API` instance call `newAPI` in the end which always sets the `information` type. So yes I think it should be ok.

sharpner:
okay, your changes sound reasonable, lets get 1.0 going and have an "unstable" API for a while. Hopefully everyone uses vendoring 😅

PR #265, opened 2016-10-14
SetID() error check

ADone:


sharpner:
nice thx for the PR - apparently golint does not work with 1.5 anymore https://github.com/golang/lint/issues/246

Issue #270, opened 2016-11-11
Is it possible to "create" associated resources at the same time as the main resource? 


carlisia:
I'm hitting a post endpoint for a resource that has a relation. To use your example, it would be much like posting a user that has data for two chocolates. 

I notice that the resource `user` gets created, but not the associated `chocolates`.

I am now doubting that this is even possible and would love it if you could confirm one way or the other. When I look at your example for a post, I see that there is no payload for the association:

https://github.com/manyminds/api2go/blob/master/examples/crud_example_test.go#L60

On my end, I've debugged enough to see that  when `SetToManyReferenceIDs` is called, the `IDs` that are passed in are correct in number (there are two associations and there are two `IDs`), but they are set to zero. This makes sense because at this point `create` hasn't been called yet.

When `create` does get called, I inspect the `obj` and the array for the association (which would be `chocolates`) is empty. If it wasn't empty I could grab the data and save it to the db.

I have implemented all the interfaces to make associations work, at least I'm pretty sure I have.

wwwdata:
Hi,

this should work, however the example is very basic and I haven't implemented that in the code. I will try to add this feature in the example and extend the example tests to verify that this really works when implemented.

carlisia:
Hey @wwwdata I'd appreciate that. Especially since the alternative is to create one resource then make individual api calls for each relation.

I think I was very diligent in making sure I was doing everything correctly, so I'm looking forward to the results of your test.

wwwdata:
I found the problem in our examples. As expected I just made a mistake. When reference IDs are in the payload, the `SetToManyReferenceIDs` and `SetToOneReferenceID` depending on the reference should be called in your resource model. Depending on what you do there, this should be working.

carlisia:
Ok I will make that change. But still, I noted that in the updated example the ID is `1` (https://github.com/manyminds/api2go/pull/271/files#diff-f6c08d9b83fa8e3e50949d71d299271dR223). But it can't be `1` in real life because it hasn't been created yet. So it will always be zero, correct?

wwwdata:
ah, now I understand what you mean. You want to create the user and new referenced data in one request. So one user and 2 chocolate datasets. Hmm. Are you trying to send the chocolates in the `included` part of the JSON? 

I think there also is no official example in the jsonapi.org spec, right? They only have examples with reference IDs.

carlisia:
That is precisely what I'm trying to do.

carlisia:
[...]

PR #271, opened 2016-11-11
Example: Create user with chocolate relationship

wwwdata:
When requesting the users, the iteration was over models and not only the reference IDs which caused a bug that no references were in the response JSON when creating a user with some initial sweets.

#270


sharpner:
ah nice catch 👍

PR #277, opened 2016-12-16
Fix fetching to-one relationships

icgood:
The `GET /:type/:id/:rel` path currently only works for to-many relationships. This PR expands it to check for to-one relationships. To query them, it looks up the reference ID from `GetReferenceIDs()` and uses `FindOne()` on the related resource to query it.


icgood:
Needs a little more work. I will re-open later.

icgood:
After looking further at the tests, I think I understand how to do what I want without upstream changes. Thank you.

PR #278, opened 2016-12-18
Allow elements to implement custom links

icgood:
If this is merged, structs may implement methods like the following:

```go
func (s MyStruct) GetCustomLinks(base string) jsonapi.CustomLinks {
    return jsonapi.CustomLinks{
        "firstlink": jsonapi.CustomLink{Href: base + `/firstlink`},
        "secondlink": jsonapi.CustomLink{
            Href: base + `/secondlink`,
            Meta: map[string]interface{}{"method": "GET"},
        },
    }
}
```

The [`"links"` object documentation](http://jsonapi.org/format/#document-links) does not specify whether or not custom links are allowed, only that additional standard links may be added in the future. However, custom links are useful for discoverability of additional services the resource may provide, when behavior extends beyond basic CRUD.

Notes:
* As the `"links"` object is not currently used when marshaling the data payload elements (only top-level and relationship `"links"` are currently used) I did not implement a merge strategy. If the `"links"` object is eventually to be populated with links (such as `"self"`) in the future, a merge strategy would have to be implemented.

### DIFF
--- a/jsonapi/data_structs.go
+++ b/jsonapi/data_structs.go
@@ -62,8 +62,8 @@ func (l *Link) UnmarshalJSON(payload []byte) error {
 		return json.Unmarshal(payload, &l.Href)
 	}
 
-	obj := make(map[string]interface{})
 	if bytes.HasPrefix(payload, objectSuffix) {
+		obj := make(map[string]interface{})
 		err := json.Unmarshal(payload, &obj)
 		if err != nil {
 			return err
--- a/jsonapi/data_structs.go
+++ b/jsonapi/data_structs.go
@@ -48,6 +48,15 @@ func (c *DataContainer) MarshalJSON() ([]byte, error) {
 	return json.Marshal(c.DataObject)
 }
 
+// CustomLink represents a custom link for return in the document.
+type CustomLink struct {
+	Href string                 `json:"href"`
+	Meta map[string]interface{} `json:"meta,omitempty"`
+}
+
+// CustomLinks contains a map of CustomLink objects as given by an element.
+type CustomLinks map[string]CustomLink
+
 // Links is a general struct for document links and relationship links.
 type Links struct {
 	Self     string `json:"self,omitempty"`
@@ -64,7 +73,7 @@ type Data struct {
 	ID            string                  `json:"id"`
 	Attributes    json.RawMessage         `json:"attributes"`
 	Relationships map[string]Relationship `json:"relationships,omitempty"`
-	Links         *Links                  `json:"links,omitempty"`
+	Links         map[string]interface{}  `json:"links,omitempty"`
 }
 
 // Relationship contains reference IDs to the related structs
--- a/jsonapi/data_structs.go
+++ b/jsonapi/data_structs.go
@@ -57,13 +57,24 @@ type Link struct {
 
 // UnmarshalJSON marshals a string value into the Href field or marshals an
 // object value into the whole struct.
-func (l Link) UnmarshalJSON(payload []byte) error {
+func (l *Link) UnmarshalJSON(payload []byte) error {
 	if bytes.HasPrefix(payload, stringSuffix) {
 		return json.Unmarshal(payload, &l.Href)
 	}
 
+	obj := make(map[string]interface{})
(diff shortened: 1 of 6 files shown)


jsonapi/data_structs.go
```diff
@@ -48,6 +48,15 @@ func (c *DataContainer) MarshalJSON() ([]byte, error) {
 	return json.Marshal(c.DataObject)
 }
 
+// CustomLink represents a custom link for return in the document.
+type CustomLink struct {
```
wwwdata: hm, a link in the `data` part can either be an object with `href` and `meta` or just a string, right? How about we make this data structure with an additional field that would be the string and then we can also change the `map[string]interface{}` to directly point to this struct. A custom marshaler/unmarshaler would be needed then

jsonapi/data_structs.go
```diff
@@ -64,7 +73,7 @@ type Data struct {
 	ID            string                  `json:"id"`
 	Attributes    json.RawMessage         `json:"attributes"`
 	Relationships map[string]Relationship `json:"relationships,omitempty"`
-	Links         *Links                  `json:"links,omitempty"`
+	Links         map[string]interface{}  `json:"links,omitempty"`
```
wwwdata: this could directly be a `map[string]CustomLink` if the marshaler/unmarshaler knows when to produce a simple string instead of an object, right?

jsonapi/data_structs.go
```diff
@@ -48,6 +48,15 @@ func (c *DataContainer) MarshalJSON() ([]byte, error) {
...
+type CustomLink struct {
+	Href string                 `json:"href"`
+	Meta map[string]interface{} `json:"meta,omitempty"`
+}
+
+// CustomLinks contains a map of CustomLink objects as given by an element.
+type CustomLinks map[string]CustomLink
+
 // Links is a general struct for document links and relationship links.
 type Links struct {
```
wwwdata: this struct could also contain a `map[string]CustomLink` right? Then we could get the standard strings and additional custom stuff. and keep the reference to this struct in the `Data` struct

jsonapi/marshal.go
```diff
@@ -204,6 +211,16 @@ func marshalData(element MarshalIdentifier, data *Data, information ServerInform
 	data.ID = element.GetID()
 	data.Type = getStructType(element)
 
+	if information != nil {
+		if customLinks, ok := element.(MarshalCustomLinks); ok {
+			data.Links = make(map[string]interface{})
+			base := getLinkBaseURL(element, information)
+			for k, v := range customLinks.GetCustomLinks(base) {
+				data.Links[k] = v
```
sharpner: i think custom links should not be allowed to overwrite automatically generated links - what do you think? At least we should add a test either way to document behaviour for this case...

icgood:
@wwwdata @sharpner Thanks for the feedback, I have made (quite a few) changes now!

I addressed the feedback directly in the commits up to and including 615c5f4455699e3b5ab9231e05689b8d47dc172f. I added a custom (un/)marshaler to allow custom links to be simple strings as well, and I added checks to prevent custom links from overwriting built-in links.

The commits after that point require additional feedback. I removed the existing `Links` struct in favor of using `map[string]Link` (aka new `Links` type) everywhere, including built-in links. If and when api2go started using built-in links for data elements I was concerned about how to "flatten" that into a `map[string]interface{}` without needing to marshal and unmarshal it. I hope that logic makes sense, as I believe the end result is much cleaner with the addition of custom links.

wwwdata:
thank you very much I like this. The one thing that's missing is a unmarshal test, then we can merge this, currently this code is not covered, see: https://coveralls.io/builds/9343018/source?filename=jsonapi%2Fdata_structs.go#L60

icgood:
@wwwdata No problem! Thanks for the library, it's pretty cool. I wrote a fairly similar one for node.js here: https://github.com/tilteng/hapi-bookshelf-jsonapi

I've beefed up the tests around marshaling/unmarshaling links. Let me know if there's anything else I can do :)

wwwdata:
thank you very much!

sharpner:
thank you @icgood :)

PR #281, opened 2016-12-29
Support for custom pagination logic

icgood:
First, adds a `Pagination` map to `api2go.Request` that extracts any pagination query parameters. E.g. `?page[test]=stuff` would contain `{"test": "stuff"}`.

Next, it removes the error if pagination fields are given and `PaginatedFindAll` is not implemented. Instead, it calls `FindAll` as normal. This allows the "supported" pagination params (page/size/offset/limit) to be used with custom pagination logic.

Finally, it adds a `Pagination` struct to `api2go.Response` that can control the pagination links returned by the regular `FindAll` method. This allows `FindAll` to return something like this:

```golang
return &api2go.Response{
    // ...,
    Pagination: api2go.Pagination{
        Next: map[string]string{"cursor": "nextpage"},  // ...?page[cursor]=nextpage
        Prev: map[string]string{"cursor": "prevpage"},  // ...?page[cursor]=prevpage
        First: map[string]string{"cursor": ""},
        // no "last" link because Last: field was not given
    },
}
```

I believe this provides very flexible usage without a lot of overhead, and should be backwards compatible with existing code. I did not like using `PaginatedFindAll` because when dealing with large datasets returning the `"last"` link is not always feasible or desired, so I tried to leave it in place while allowing me to control pagination as desired.

### DIFF
--- a/api.go
+++ b/api.go
@@ -427,25 +427,24 @@ func (res *resource) marshalResponse(resp interface{}, w http.ResponseWriter, st
 }
 
 func (res *resource) handleIndex(c APIContexter, w http.ResponseWriter, r *http.Request, info information) error {
-	pagination := newPaginationQueryParams(r)
-	if pagination.isValid() {
-		source, ok := res.source.(PaginatedFindAll)
-		if !ok {
-			return NewHTTPError(nil, "Resource does not implement the PaginatedFindAll interface", http.StatusNotFound)
-		}
+	if source, ok := res.source.(PaginatedFindAll); ok {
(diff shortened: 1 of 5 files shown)


icgood:
FYI I'm not sure why Travis is failing, the error seems unrelated to my change:

```
The command "bash scripts/fmtpolice" exited with 1.
```

wwwdata:
Travis fails because of our lint checker. This was in the Travis output: 

> $ bash scripts/fmtpolice
response.go:46:9: if block ends with a return statement, so drop this else and outdent its block

Issue #267, opened 2016-10-20
Make it possible to implement only a subset of the CRUD interface


freshtonic:
Not all HTTP verbs and routes are relevant to all resources.

If I submit a PR for an alternative `addResource` function that introspects the resource for a subset if the `CRUD` methods and generates only the corresponding routes is this something that could be merged?  I'm happy to do the work.

wwwdata:
Hi. Of course that would be a nice feature.

timbunce:
Any news on this @freshtonic?

philippecarle:
That would be a great feature. 👍
But imho it would introduce a huge BC break and so it definitely needs to be developed and merged before 1.0.

philippecarle:
+ i would be up to work on that but being a quite new gopher, I think it'd be safer if i got some advice on how it should be implemented. A new `Operations` interface with a getter returning a slice of string (defined in constants) ?

wwwdata:
@philippecarle I guess the best way to split up the CRUD Operations is to split up the CRUD interface into smaller ones. So one Interface for each method. The API then has to check what interfaces are implemented and only then provides the endpoints that are really needed.

wwwdata:
This could also be done in a backwards compatible way. The CRUD interface can stay the same and consists of all the individual ones.

```go
type FindOne interface {
	FindOne(ID string, req Request) (Responder, error)
}

type CRUD interface {
	FindOne
	Create
	Delete
	Update
}
```

maybe with different interface names.

philippecarle:
I like this last option. I will probably dig into the code later in the day to find out how hard (or easy) it would be to make this work (and most of all how to update the tests to cover that).

freshtonic:
@timbunce I still think this would be a great feature but I'm no longer involved with the project that would have benefited from this and I no longer have the time to contribute.

All the best!

PR #286, opened 2017-06-04
Split CRUD interface

philippecarle:
Here is a first attempt to split the CRUD interface as mentioned in issue #267.

I really don't like typing the source with an empty interface but as the idea here is to be able to pass one implement of at least one of the new four interfaces, I can't see any other way to make it work.

Suprisingly, tests are passing seamlessly. I probably say surprisingly because I come from a inheritance-based language and composition in Go allows us to make smooth updates on interfaces (❤️). By the way, we do need to test the generated routes by passing several combinations of implementations of these CRUD interfaces.

Let me know what do you think about that !

### DIFF
--- a/api.go
+++ b/api.go
@@ -266,7 +266,7 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source interfac
 		}
 	})
 
-	if _, ok := source.(CRUDFindOne); ok {
+	if _, ok := source.(ResourceGetter); ok {
 		api.router.Handle("OPTIONS", baseURL+"/:id", func(w http.ResponseWriter, r *http.Request, _ map[string]string) {
 			c := api.contextPool.Get().(APIContexter)
 			c.Reset()
@@ -366,7 +366,7 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source interfac
 		}
 	}
 
-	if _, ok := source.(CRUDCreate); ok {
+	if _, ok := source.(ResourceCreator); ok {
 		api.router.Handle("POST", baseURL, func(w http.ResponseWriter, r *http.Request, params map[string]string) {
 			info := requestInfo(r, api)
 			c := api.contextPool.Get().(APIContexter)
@@ -380,7 +380,7 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source interfac
 		})
 	}
 
-	if _, ok := source.(CRUDDelete); ok {
+	if _, ok := source.(ResourceDeleter); ok {
 		api.router.Handle("DELETE", baseURL+"/:id", func(w http.ResponseWriter, r *http.Request, params map[string]string) {
 			c := api.contextPool.Get().(APIContexter)
 			c.Reset()
@@ -393,7 +393,7 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source interfac
 		})
 	}
 
-	if _, ok := source.(CRUDUpdate); ok {
+	if _, ok := source.(ResourceUpdater); ok {
 		api.router.Handle("PATCH", baseURL+"/:id", func(w http.ResponseWriter, r *http.Request, params map[string]string) {
 			info := requestInfo(r, api)
 			c := api.contextPool.Get().(APIContexter)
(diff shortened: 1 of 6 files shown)


api.go
```diff
@@ -364,40 +366,46 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD) *r
...
-		c := api.contextPool.Get().(APIContexter)
-		c.Reset()
-		api.middlewareChain(c, w, r)
-		err := res.handleCreate(c, w, r, info.prefix, *info)
-		api.contextPool.Put(c)
-		if err != nil {
-			api.handleError(err, w, r)
-		}
-	})
+	if _, ok := source.(CRUDCreate); ok {
```
sharpner: I kinda stumble across the interface names, I know that you just tried to keep it consistent. But maybe CRUD wasn't the best name either. :P 

How about something like:
`ResourceCreator` `ResourceUpdater` or even
`SingleResourceCreator` 

I'm so bad at naming things but imho the interface name should not simply be the name of its method

philippecarle: I knew that naming would be a thing here. I wasn't much creative and just tried to keep a reference to the original interface. Your naming makes sense: it describes an action applied to a resource.... committing an update right now ;)

api.go
```diff
@@ -275,17 +266,28 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source CRUD) *r
...
-		if err != nil {
-			api.handleError(err, w, r)
-		}
-	})
+	if _, ok := source.(CRUDFindOne); ok {
+		api.router.Handle("OPTIONS", baseURL+"/:id", func(w http.ResponseWriter, r *http.Request, _ map[string]string) {
+			c := api.contextPool.Get().(APIContexter)
+			c.Reset()
+			api.middlewareChain(c, w, r)
+			w.Header().Set("Allow", "GET,PATCH,DELETE,OPTIONS")
```
philippecarle: It sounds logical and I'm working on it. Kind of hard not to duplicate the code for this one and the `"GET,PATCH,DELETE,OPTIONS"` a few lines below.

philippecarle:
I'll be writing the tests asap. Probably in the next 2 hours 😉

api.go
sharpner: you could implement a function `getAllowedMethodsByResource()` which creates the string based on the implemented interfaces, and use this function on both places - but maybe you'll find another solution to this :)

philippecarle: Actually there are two strings: one for the collection and another one for the item...
Here is the full logic, written very procedurally:
```go
allowedForCollection := [4]string{}
	allowedForItem := [4]string{}

	allowedForCollection[0] = http.MethodOptions
	allowedForItem[0] = http.MethodOptions

	if _, ok := source.(CRUDFindOne); ok {
		allowedForCollection[1] = http.MethodGet
		allowedForItem[1] = http.MethodGet
	}

	if _, ok := source.(CRUDCreate); ok {
		allowedForCollection[2] = http.MethodPost
	}

	if _, ok := source.(CRUDUpdate); ok {
		allowedForCollection[3] = http.MethodPatch
		allowedForItem[2] = http.MethodPatch
	}

	if _, ok := source.(CRUDDelete); ok {
		allowedForItem[3] = http.MethodDelete
	}
```

philippecarle: Fixed in a method with a boolean `collection` argument

api.go
philippecarle: Naming improved

philippecarle:
Tests written (the KISS way, as most tests should be, but could probably have been improved/refactored, let me know what do you think about them).
I even updated the README 😜

philippecarle:
Just a suggestion : maybe we should keep the original CRUD interface embedding ˋResourceCreator`, ˋResourceUpdater` and ˋResourceDeleter` in order that one can implement everything at once easily with the IDE ?

sharpner:
ah yeah sure! It makes no sense declaring every method twice :)

philippecarle:
The "main" interface is back, README is updated too. I named it `ResourceCRUD` to make it consistent with the other ones, wdyt ?
I may be opening other PRs on this project as I will be working intensively with it for a personal project in the next weeks. I'll probably open some issues to debate about ideas I have (especially one about authorization) and to make sure the submitted implementations would match your requirements / ideas about api2go.

sharpner:
Since renaming the interface would introduce a BC break, I'd rather keep the old name for now, even though the name one is actually better. Don't hesistate to open up PRs - they are always welcome :)

philippecarle:
You nailed it! Just squashed the update with the previous commit and force pushed.

wwwdata:
thank you very much @philippecarle 
more PRs and feedback/improvements are always very appreciated!

PR #285, opened 2017-06-02
Added support for custom meta info in relations.

michaelkleinhenz:
This adds custom meta support for relations:
```
"relationships": {
  "author": {
    "links": {
      "self": "http://my.domain/v1/posts/someID/relationships/author",
      "related": "http://my.domain/v1/posts/someID/author"
    },
    "meta": {
      "someMetaKey": "someMetaValue",
      "someOtherMetaKey": "someOtherMetaValue"
    }
  }
}
```
To use this feature, a type has to implement `MarshalCustomRelationshipMeta` and return the meta structure for the type's relations from `GetCustomMeta()`:
```
type MarshalCustomRelationshipMeta interface {
	MarshalIdentifier
	GetCustomMeta(string) Metas
}
```
A test is included.

PR #291, opened 2017-10-04
Custom meta in relation

sharpner:
this PR continues #285


sharpner:
This is a BC break, should we create a tag before merging?

Issue #292, opened 2017-10-12
httprouter.go Need Type Assertion


Linuturk:
Hey guys,

Not sure why I'm getting this error when I attempt to build with `api2go`:

```
github.com/manyminds/api2go/routing/httprouter.go:51: cannot use notAllowedHandler (type http.Handler) as type http.HandlerFunc in assignment: need type assertion
```

Seems to be this line here: https://github.com/manyminds/api2go/blob/master/routing/httprouter.go#L51

If I comment this out everything seems to validate fine. Given the code in that section is from two years ago, I imagine there's just a small tweak that needs to be made?

sharpner:
Can you show us how you are using it? A bit of code might help to find the problem.

sharpner:
Tried to dig into this blindly, but I can't reproduce any error:
```
package main
import (
        "fmt"
        "net/http"
        "github.com/manyminds/api2go"
        "github.com/manyminds/api2go/routing"
)
func test() http.Handler {
        return nil
}
func main() {
        routing.NewHTTPRouter("a", test())
        test := api2go.NewAPIWithBaseURL("/", "/")
        fmt.Printf("%'v", test)
}
``` 
works?

Linuturk:
Yeah, sorry for the lack of context.

```
$ go version
go version go1.9.1 linux/amd64
```

Here are the related sections of `dep status`, just updated yesterday before submitting the issue.
```
github.com/gorilla/context           *              v1.1           1ea2538   1ea2538  1  
github.com/gorilla/handlers          ^1.2.1         v1.2.1         a4043c6   a4043c6  1  
github.com/gorilla/mux               ^1.4.0         v1.5.0         24fca30   24fca30  1
github.com/manyminds/api2go          *              0.6            02704db   02704db  3 
github.com/manyminds/api2go-adapter  *              branch master  13346ba   13346ba  1
```

Was trying to figure out how to integrate into my existing code base that uses gorilla mux. I found this Example and tried to go down this route: https://github.com/manyminds/api2go-adapter/tree/master/gorillamux

```
	r := mux.NewRouter()
	api := api2go.NewAPIWithRouting(
		"api",
		api2go.NewStaticResolver("/"),
		api2go.DefaultContentMarshalers,
		gorillamux.New(r),
	)
... r.HandleFunc routes here ...
... LIstenAndServe...
```

sharpner:
so it is fixed for you?

Linuturk:
[...]

PR #293, opened 2017-10-13
httprouter.go Need Type Assertion

Linuturk:
Per https://github.com/manyminds/api2go-adapter/issues/5 and discussion in https://github.com/manyminds/api2go/issues/292


Linuturk:
I don't seem to have a good enough understanding of why this isn't working in my situation, and this "fix" seems to break a lot more.

Issue #290, opened 2017-10-04
How to add multiple API versions


imhuytq:
What solution can I add more than one version of the API? Examples v1, v1.2, v2 ...

sharpner:
Okay, this does not 100% work as intended, but you can mux it with a custom handler:

```
package main
import (
        "fmt"
        "log"
        "net/http"
        "strings"
        "github.com/manyminds/api2go"
        "github.com/manyminds/api2go/examples/model"
        "github.com/manyminds/api2go/examples/resource"
        "github.com/manyminds/api2go/examples/storage"
)
func main() {
        v1 := api2go.NewAPI("v1")
        v2 := api2go.NewAPI("v2")
        userStorage := storage.NewUserStorage()
        chocStorage := storage.NewChocolateStorage()
        v1.AddResource(model.User{}, resource.UserResource{ChocStorage: chocStorage, UserStorage: userStorage})
        v2.AddResource(model.User{}, resource.UserResource{ChocStorage: chocStorage, UserStorage: userStorage})
        handler := http.DefaultServeMux
        handler.HandleFunc("/", func(rw http.ResponseWriter, r *http.Request) {
                log.Printf("%s %b\n", r.RequestURI, strings.HasPrefix("/v1", r.RequestURI))
                if strings.HasPrefix(r.RequestURI, "/v1") {
                        v1.Handler().ServeHTTP(rw, r)
                        return
                }
                v2.Handler().ServeHTTP(rw, r)
                return
        })
        http.ListenAndServe(fmt.Sprintf(":5050"), handler)
}
```

The prefered solution would be, to register both API versions directly at the mux.... @wwwdata got any ideas on how to solve this?

Using this snippet:
```
        v1 := api2go.NewAPI("v1")
        v2 := api2go.NewAPI("v2")
        userStorage := storage.NewUserStorage()
        chocStorage := storage.NewChocolateStorage()
        v1.AddResource(model.User{}, resource.UserResource{ChocStorage: chocStorage, UserStorage: userStorage})
        v2.AddResource(model.User{}, resource.UserResource{ChocStorage: chocStorage, UserStorage: userStorage})
        handler := http.DefaultServeMux
        handler.Handle("/", v1.Handler())
        handler.Handle("/", v2.Handler())
        http.ListenAndServe(fmt.Sprintf(":5050"), handler)
```

will of course yield:
```panic: http: multiple registrations for /```

wwwdata:
[...]

PR #295, opened 2017-10-14
Add easy way to combine multiple API versions

wwwdata:
Make it easy to add another version of an API to the routing of the first. This is fixing #290


sharpner:
wow that was easy ;)

PR #310, opened 2018-01-24
Bugfix lb 333 escape comma in url params

mukundnathan:
String with commas will be split as per api2go documentation
"If there are multiple values, you have to separate them with a comma. api2go automatically slices the values for you"
Hence urlencode comma with escape if commas has to be treated as inclusive in string
"hello, world" should be encoded "hello\, world" or "hello%5c%2c world".


mukundnathan:
not for this repo sorry

PR #313, opened 2018-04-18
Support links with null values

ryanuber:
I have an API which returns pagination links (`prev`, `next`, etc.) with null values if there is no previous or next page. I am using this library to write a Go client for it and noticed I started getting `expected a JSON encoded string or object` errors. At first I thought our API was wrong, but after a quick check of the spec, it turns out this actually is something JSONAPI-compliant clients should support.

On first read, the spec seems to contradict itself a bit. Here is bit from the [links element](http://jsonapi.org/format/#document-links), which states:

> The value of each links member MUST be an object (a “links object”).

And here is a quote from the [pagination section](http://jsonapi.org/format/#fetching-pagination):

> Keys MUST either be omitted or have a null value to indicate that a particular link is unavailable.

I think the best thing to do is probably to allow null link values by simply ignoring them, which is what this PR does. `Unmarshal` code paths will be able to simply check the `Links` object for presence of the link they are looking for, and `Marshal` code paths will continue to work. The only difference here would be that null links would simply not be included in the marshaled output, instead of returning a link object with an empty `href` and no `meta`.

### DIFF
--- a/jsonapi/data_structs.go
+++ b/jsonapi/data_structs.go
@@ -58,6 +58,12 @@ type Link struct {
 // UnmarshalJSON marshals a string value into the Href field or marshals an
 // object value into the whole struct.
 func (l *Link) UnmarshalJSON(payload []byte) error {
+	// Links may be null in certain cases, mainly noted in the JSONAPI spec
+	// with pagination links (http://jsonapi.org/format/#fetching-pagination).
+	if len(payload) == 4 && string(payload) == "null" {
+		return nil
+	}
+
 	if bytes.HasPrefix(payload, stringSuffix) {
 		return json.Unmarshal(payload, &l.Href)
 	}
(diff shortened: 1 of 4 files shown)


sharpner:
Thanks for the thorough investigation :) 

You are right we should support null values for links, as currently both is possible to be implemented on the backend site. 
I opened a PR to fix the build, since the failure has nothing to do with this PR.

I just not sure if it's a good idea to silently drop the link keys if their value is null, any pitch for doing so?

ryanuber:
@sharpner I think dropping the links during `Marshal` is decent, because you otherwise end up with some strange JSON output such as:
```
...
"links": {
  "foo": { "href": "", "meta": {}},
...
```

As for the `Unmarshal` dropping them, we could opt to _not_ drop the link names from the result. I just found it a weird API to try checking if a link was empty or not because we are dealing with struct values, not pointers, so the `Link` cannot be `nil`. Perhaps making `(l Link) Empty() bool` a public method would make it nicer to use. Personally I would sooner just check if the link I want is present in the `Links`. The only valid use case I know of for null links is to communicate that a given link is unavailable. Let me know what you think/prefer.

ryanuber:
@sharpner I gave preserving the null link values a rip and that seems to work out pretty well and be a tad less complicated. I pushed a change up - let me know how that sits with you.

PR #320, opened 2018-09-29
Allow resources to specify meta using the new MarshalMeta interface

L-four:
As per the [jsonapi](http://jsonapi.org/format/#document-resource-objects) Resource Objects should be able to have "meta" member pointing to a "Meta Object"


L-four:
I added a basic test and updated the method as per @sharpner recommendation but now the travis-ci build/tests are failing  it appears that "github.com/golang/lint/golint" has moved or am I missing something?

wwwdata:
thanks, can you rebase on master, then the travis tests should be running again

L-four:
Thanks for resolving the Travis-ci issue rebase on master should be good to go now. 😄

PR #324, opened 2019-03-24
Adding RouteNamer interface

mikeschinkel:
Using api2go I noticed you coupled the types with the route names, and I want to have singular types and plural routes. So I added a `RouteNamer` interface.  If you are willing to accept this I will be happy to update the docs too.

### DIFF
--- a/api.go
+++ b/api.go
@@ -244,6 +244,12 @@ func (api *API) addResource(prototype jsonapi.MarshalIdentifier, source interfac
 		return info
 	}
 
+	// check if EntityNamer interface is implemented and use that as name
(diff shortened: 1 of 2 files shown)


wwwdata:
Thanks, LGTM but before we merge we should get the tests back to be running. I will look into this, merge it in and then you can rebase on master.

wwwdata:
Ok, fixed the build problems in PR #325 

What you could also do is add a small test to check that this functionality works so that we can guarantee that nothing breaks it in the future. Then we rebase after merging my PR and can merge this.

wwwdata:
@mikeschinkel ping: would you like to continue here? otherwise I can also take over and merge after adding tests

PR #325, opened 2019-03-24
Fix travis build

wwwdata:
Make travis run again.

The echo framework already upgraded to the new module system. We should also look into that and support it.


sharpner:
thx 👍

Issue #328, opened 2019-04-29
Errors from SetToOneReferenceID and SetToManyReferenceIDs are ignored


velovix:
The SetToOneReferenceID and SetToManyReferenceIDs methods are defined as returning an error value, but that value is ignored in the places they are called. If these errors were to simply propagate up the stack to the Marshal and Unmarshal calls, I think that would be sufficient.

wwwdata:
Would you like to fix this and open a PR?

velovix:
I'd be happy to! I'll see what I can do.

velovix:
Resolved by #329

PR #329, opened 2019-05-10
Add missing error checks to SetToOneReferenceID and SetToManyReferenceIDs calls

velovix:
This PR fixes some instances where these methods are being called but their errors are not being checked. Also adds a test in the jsonapi library for this behavior and fixes the behavior of a fixture that used to return an error that was ignored.


velovix:
Resolves #328

velovix:
Sorry, I accidentally included the module changes in this branch and had to remove them.

PR #330, opened 2019-05-10
Add module support

velovix:
This PR turns api2go into a module. Modules are Go's new, official system for dependency management! The feature is currently experimental but it will very likely be finalized in Go 1.13 and the Go team has committed to making sure modules made today will work in Go 1.13 and beyond. Older versions of Go that don't support modules will simply ignore these files and do what they've always done.


velovix:
Anything I can do to help this land?

PR #336, opened 2020-10-12
Add local id support

IvanRN:
According to upcoming v1.1 of the JSONAPI the document object
identification is a combination of three values: id, type and
lid (local id). Local id is an id that can be generated when
entity was created but not yet persisted to the DB. In the
JSONAPI it should allow client to match data which he created
and then received back from the server.


sharpner:
maybe next time :P

PR #352, opened 2023-05-04
Introduce LocalID and update go.mod dependencies to the latest

artlo:
Introducing the LocalID feature that was introduced in jsonapi specification.
- LocalID is added to ID, single and multiple Reference ID
- GetID method was reworked that it returns the identifier object. The identifier contains the ID, LID and model Name
- remove GetName method since the Name is part of Identifier

Update the dependencies to the latest and as result have to support only the latest Go versions.


artlo:
@wwwdata , Could you please take a look if you have a time?

